PR #29, opened 2018-09-17
Adding tip on adding bin folder to PATH

sublibra:
Simplifying for more plug and play based based audience. 

Cluttering or supporting? You decide..

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,10 @@ mkdir -p "$HOME/bin" &&
 curl -L https://github.com/qlik-oss/corectl/releases/download/v0.0.2/corectl-Windows-x86_64.exe > "$HOME/bin/corectl.exe" &&
 chmod +x "$HOME/bin/corectl.exe"
 ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -28,6 +28,10 @@ mkdir -p "$HOME/bin" &&
 curl -L https://github.com/qlik-oss/corectl/releases/download/v0.0.2/corectl-Windows-x86_64.exe > "$HOME/bin/corectl.exe" &&
 chmod +x "$HOME/bin/corectl.exe"
 ```
+Optionally you can add the bin to the PATH environment variable to simplify usage in the command line.
+```bash
+setx PATH "%PATH%;%HOME%/bin"
```
wennmo: I'm a bit hesitant to tips like this unless it is something needed specifically for this tool. The problem is often that the commands are OS specific and might require elevated privileges.

sublibra: Agree. I am leaning to more cluttering of the docs.

Issue #11, opened 2018-08-30
Dependency to local running docker


carlioth:
At startup there is a dependency of having an engine running on 127.0.0.1:9076

Handle this in a better way that just displaying: 
` Could not connect to engine:localhost dial tcp 127.0.0.1:9076: connect: connection refused `

PR #38, opened 2018-09-19
Better error message when not selecting an engine

FredrikFolkesson:
For issue #11 

Just giving a more clear error message when no engine is specified (and not defaulting to localhost)

### DIFF
--- a/main.go
+++ b/main.go
@@ -64,6 +64,11 @@ var (
 			ttl := viper.GetString("ttl")
 			sessionID := "QLISession-" + appID
 
+			if engine == "" {
+				fmt.Println("No engine specified.")
+				fmt.Println("Specify using the --engine parameter or in your config file")
+				os.Exit(0)
+			}
 			if appID == "" {
 				fmt.Println("Using session app")
 			}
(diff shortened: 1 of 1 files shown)


wennmo:
I think we should add this information to the README as well

hrigner:
If we add it to the readme as @wennmo suggests, then we should add info about all the other config parameters as well?
IMO, "Specify what Qlik Associative Engine to use with the --engine parameter" is enough

wennmo:
I think it needs to be clearer in the documentation what is required and optional when running the tool. If we are defaulting to values if the parameter is not set, it should also be part of the documentation imo.

carlioth:
I think the error message is fine, but I would like us to print the usage after. gcloud is a good example. Example attached
screen shot 2018-09-20 at 08 54 23

FredrikFolkesson:
I agree that the docs needs to be updated, but that goes for a lot of the parameters, different issue?

FredrikFolkesson:
@carlioth instead or in addition to what is currently being printed?

carlioth:
@FredrikFolkesson In addition to the error message. 
<Error message>
<Usage printout>

Issue #30, opened 2018-09-17
Running the tool agains Qlik Sense creates an unnamed-app in the engine


sublibra:
The tool should not leave artifacts in engine after running meta commands etc.

Issue #17, opened 2018-09-03
Require app to have data loaded on majority of commands


peol:
Commands in bold below should probably require an existing app, and data loaded in it, to work (i.e. print error and quit):

* *assoc*
* *eval*
* *field*
* fields
* help
* *keys*
* meta
* reload
* script
* *tables*

PR #39, opened 2018-09-19
Improved error handling and a bonus status command

### LINKED ISSUE #34: Handle using eval without call to meta/reload/...
Corectl should handle calling on eval when no other call to meta reload or other command has been used.

gabbaxx:
Fixes #34, #30, #17
- Added a check that there is at least one table in all view commands (fields, eval etc)
- Made session id more unique (but still persistent) not to clash if more than one user accesses the same engine
- No more accidental creation of unnamed apps. If no app is specified a session app is used.
- Added a new status command which sometimes is needed to understand issues where corectl is accidentally connected to the wrong engine or wrong app (for instance when pointing at the wrong config file).

### DIFF
--- a/internal/modeldata.go
+++ b/internal/modeldata.go
@@ -3,6 +3,7 @@ package internal
 import (
 	"context"
 	"fmt"
+	"log"
 	"os"
 
 	"github.com/qlik-oss/enigma-go"
@@ -121,12 +122,10 @@ func addTableFieldCellCrossReferences(fields []*FieldModel, tables []*TableModel
 func GetModelMetadata(ctx context.Context, doc *enigma.Doc, metaURL string, keyOnly bool) *ModelMetadata {
 	tables, sourceKeys, err := doc.GetTablesAndKeys(ctx, &enigma.Size{}, &enigma.Size{}, 0, false, false)
 	if err != nil {
-		fmt.Println(err)
-		os.Exit(1)
+		log.Fatalln(err)
 	}
 	if len(tables) == 0 {
-		fmt.Println("The data model is empty.")
-		os.Exit(1)
+		log.Fatalln("The data model is empty.")
 	}
 	restMetadata, err := ReadRestMetadata(metaURL)
 
@@ -195,11 +194,9 @@ type FieldSourceTableInfo struct {
 func ensureModelExists(ctx context.Context, doc *enigma.Doc) {
 	tables, _, err := doc.GetTablesAndKeys(ctx, &enigma.Size{}, &enigma.Size{}, 0, false, false)
 	if err != nil {
-		fmt.Println(err)
-		os.Exit(1)
+		log.Fatalln(err)
 	}
 	if len(tables) == 0 {
-		fmt.Println("The data model is empty.")
-		os.Exit(1)
+		log.Fatalln("The data model is empty.")
 	}
 }
--- a/internal/modeldata.go
+++ b/internal/modeldata.go
@@ -3,8 +3,9 @@ package internal
 import (
 	"context"
 	"fmt"
-	"github.com/qlik-oss/enigma-go"
 	"os"
+
+	"github.com/qlik-oss/enigma-go"
 )
 
 // ModelMetadata defines all available metadata around the data model.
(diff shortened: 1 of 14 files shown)


test/corectl_integration_test.go
```diff
@@ -92,6 +92,9 @@ func TestCorectl(t *testing.T) {
 
 		{"reload project 2", []string{"--config=test/project2/qli.yml ", connectToEngine, "reload"}, "project2-reload.golden"},
 		{"fields project 2", []string{"--config=test/project2/qli.yml ", connectToEngine, "fields"}, "project2-fields.golden"},
+
+		{"reload project 2", []string{"--config=test/project3/qli.yml ", connectToEngine, "reload"}, "project3-reload.golden"},
```
wennmo: wrong project number

internal/modeldata.go
```diff
@@ -185,3 +189,16 @@ type FieldSourceTableInfo struct {
 	CellContent string
 	KeyType     string
 }
+
+// Exists if there  is no data model
```
wennmo: typo

main.go
```diff
@@ -228,7 +253,7 @@ func init() {
 	corectlCommand.PersistentFlags().String("engine-headers", "30", "HTTP headers to send to the engine")
 	viper.BindPFlag("engine-headers", corectlCommand.PersistentFlags().Lookup("engine-headers"))
 
-	corectlCommand.PersistentFlags().StringP("app", "a", "unnamed-app.qvf", "App name including .qvf file ending")
+	corectlCommand.PersistentFlags().StringP("app", "a", "", "App name including .qvf file ending")
```
wennmo: I think this part needs to be documented in the README as well, since the documentation is quite unclear right now

printer/status.go
```diff
@@ -0,0 +1,37 @@
...
+		fmt.Println("The data model is empty.")
+	} else if tableCount == 1 {
+		fmt.Println("The data model has 1 table.")
+	} else {
+		fmt.Printf("The data model has %d tables.", tableCount)
+	}
+
+}
+
+//Returns the number of
```
wennmo: the number of tables in data model?

internal/state.go
```diff
@@ -49,32 +47,28 @@ func PrepareEngineState(ctx context.Context, engine string, sessionID string, ap
...
-			appID := "SessionDoc"
-			LogVerbose("Document: " + appID + "(reconnected)")
+	doc, err = global.GetActiveDoc(ctx)
+	if doc != nil {
+		// There is an already opened doc!
+		if appID != "" {
+			LogVerbose("App: " + appID + "(reconnected)")
 		} else {
-			LogVerbose("No active doc: " + appID + ", session=" + sessionID)
+			LogVerbose("Session app (reconnected)")
```
wennmo: We should probably add the `sessionID` to the log print here, so it's clear which session it reconnects to.

wennmo: The session id is already part of an earlier log event, so no need to add it here.

main.go
wennmo: Done

printer/status.go
```diff
@@ -0,0 +1,37 @@
...
+func PrintStatus(state *internal.State) {
+	if state.AppID != "" {
+		fmt.Println("Connected to " + state.AppID + " @ " + viper.GetString("engine"))
+	} else {
+		fmt.Println("Connected to session app @ " + viper.GetString("engine"))
+	}
+	tableCount := dataModelTableCount(state.Ctx, state.Doc)
+	if tableCount == 0 {
+		fmt.Println("The data model is empty.")
+	} else if tableCount == 1 {
```
FredrikFolkesson: change to "The data model has %d table(s.) and skip one if-else?

internal/state.go
```diff
@@ -125,6 +119,9 @@ func buildWebSocketURL(engine string, ttl string) string {
 }
 
 func buildMetadataURL(engine string, appID string) string {
+	if appID == "" {
```
FredrikFolkesson: Should appId ever be an empty string? and can't you get metadata for a sessionApp?

PR #42, opened 2018-09-20
Update connections file target, integrationtest to show without config.

carlioth:
Update to have the same name both doc and implementation for connections file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 vendor/
 .DS_Store
 *.qvf
+corectl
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -2,3 +2,4 @@
 vendor/
 .DS_Store
 *.qvf
+corectl
```
wennmo: What about the windows users? 😉

carlioth: windows what :) , I fix

PR #44, opened 2018-09-21
Adding command to list apps

hrigner:
Purposely not adding any info to readme since it was decided to have the --help output there

### DIFF
--- a/main.go
+++ b/main.go
@@ -234,6 +234,23 @@ var (
 			printer.PrintStatus(state)
 		},
 	}
+
+	listAppsCmd = &cobra.Command{
+		Use:   "apps",
+		Short: "Print app list",
+		Long:  "Print app list",
+
+		Run: func(ccmd *cobra.Command, args []string) {
+			docList, err := state.Global.GetDocList(state.Ctx)
+
+			if err != nil {
+				fmt.Println(err)
+				os.Exit(-1)
+			}
+
+			printer.PrintApps(docList)
+		},
+	}
 )
 
 func getSessionID(appID string) string {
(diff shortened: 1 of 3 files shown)


hrigner:
@carlioth, is the output satisfying?

carlioth:
@hrigner will try the branch now!

main.go
```diff
@@ -234,6 +234,23 @@ var (
 			printer.PrintStatus(state)
 		},
 	}
+
+	listAppsCmd = &cobra.Command{
+		Use:   "apps",
```
FredrikFolkesson: Just checking because I dont know.

Is it apps or is it docs? And are we consistent?

main.go
```diff
@@ -234,6 +234,23 @@ var (
...
+		Short: "Print app list",
+		Long:  "Print app list",
+
+		Run: func(ccmd *cobra.Command, args []string) {
+			docList, err := state.Global.GetDocList(state.Ctx)
+
+			if err != nil {
+				fmt.Println(err)
```
FredrikFolkesson: use log.fataln(err)

hrigner:
The format is enigma go Float64, which basically is a float64 (Float64 is an enigma-go equivalent of float64 which adds support for the Qlik Associative Engine specific way of marshalling and unmarshalling "Infinity", "-Infinity" and "NaN" as json strings. It can always safely be typecasted to plain float64 including these special cases)

 Maybe we should convert is to something more readable?

main.go
hrigner: It's apps nowadays.. but it used to be docs :)

carlioth:
Would be great if we could get it into a readable format when printing, or else I think we should remove that field from the print.

PR #51, opened 2018-09-21
Generate markdown from cobra command

### LINKED ISSUE #40: Document the different command line parameters in the readme
Which parameters exists
What do they do
Which ones are required
What are the default values for the ones that are not required to be set

### LINKED ISSUE #46: Remove tool help part form readme.md and replace with an automatic help output printout in readme.

wennmo:
Cobra has built-in functionality for generating markdowns from commands. I added a command to generate the docs: `corectl generate-docs`. This command  is hidden from the rest and is not part of the help output. 

Also added a check in Circle CI to check if the documentation should be updated.

There is still a need to add better descriptions to the different commands, but this way we can generate the documentation and keep it up-to-date.

This closes #40 and closes #46


wennmo:
@peol I agree, however the support for generating the markdowns is quite limited in options. Might require that we implement handling of already generated markdowns if we want it to look a certain way.

@FredrikFolkesson The examples should be documented in each command imo, but I agree that they are needed :)

PR #65, opened 2018-10-05
First version attempt

### LINKED ISSUE #49: Add -v / --version flag
Add the flag -v and --version and return the version of the corectl tool.

FredrikFolkesson:
This is one way to solve versioning in corectl.

Only supports `corectl version` since the `-v` flag means verbose (same as kubectl)

The version string will be added during compile time like this:
`go build -ldflags "-X main.version=`cat version.txt`"`

For releases, I am thinking that you have a script that you run locally `release.sh minor` (in the same way as for mira for example)

That will change the version.txt to match the version to be released and pushes a git tag along with the commit. CircleCi will have a build step that matches on the git-tag and does the actual building of corectl for our different OS:es (if possible?) and then either creating a github release directly with the files uploaded or saving them somewhere so we can do a github release ourselves.

In the future we could also add stuff like OS/build time/commit sha and the likes. But I think we should start with the version number and add the rest if/when we need it.

This closes #49


version.txt
```diff
@@ -0,0 +1 @@
+0.0.4
```
peol: Should be 0.0.3 since that's the current release, no?

FredrikFolkesson: No, I am thinking that the version.txt holds the version to be released. (I.E the one we are making prereleases on now)

So if we would choose to produce artifacts on our master commits they would be versioned 0.0.4-buildnumber because they are pre-releases of version 0.0.4 (and 0.0.3-buildnumber does not make any sense since version 0.0.3 is already released)

And this is how we do in mira and I think comes from cloud

But If we want to have it the otherway that is fine by me as well :)

peol: I personally prefer to keep the version master being the latest release, and instead tag the rolling builds with like `20181004-<buildnumber>` or something similar if needed. You don't know after releasing a version what the next version will be so it doesn't make sense to me that you'd bump it prematurely. :)

PR #77, opened 2018-10-15
adding release flow with goreleaser

qlikossbuild:
Using goreleaser for releases.

A release is initiated by pushing a tag on the format `v.*` so `v0.0.4` for example (we could make this stricter if we want)

The version of the produced program will be the value of the tag. So the `version` file is removed.
(We could keep it just to show the latest version if you have pulled the repo, or to use if we would add a script instead of manually pushing tags, but I don't think it is worth it)

Goreleaser will publish a GitHub release with the built binaries for the specified tag.


FredrikFolkesson:
I forgot to logout from the bot before creating this PR

.goreleaser.yml
```diff
@@ -0,0 +1,28 @@
...
+    - linux
+    - windows
+  goarch:
+    - amd64
+
+before:
+  hooks:
+    - dep ensure
+
+archive:
```
wennmo: Will a github release only contain the binaries?

carlioth: As I understood it, notes of sha265 for the files as well as the commit history will be produced as notes when releasing.

carlioth: Regarding the readme bump, I think it would be nice with a "correct" URL to make it easier to download. @FredrikFolkesson is there any way you can hook into a release and add this?

FredrikFolkesson: The github release will be 3 zipped files (which only includes the binaries)

SHA:s and changelog should as I understand it be added to the release itself.

@carlioth 
There does not seem to be a way to link to a artifact from "latest" release :( Since we are only pushing tags I dont think we can do any changes.

I think changing the wording to <version> or something should be good enough.

PR #79, opened 2018-10-16
Remove global flags from where not applicable. Fixes #66

### LINKED ISSUE #66: Remove global flags if they are not global
Currently, most flags should not be global (They dont apply to the `version` command for example)

They should be moved to the commands where they apply

gabbaxx:


wennmo:
I agree with @FredrikFolkesson. This change does not remove them from the doc, which is quite confusing.

wennmo:
Also the `Fixes #nbr` should be part of the PR description and not the title. Otherwise it does not link to the issue.

FredrikFolkesson:
closed with #96

PR #84, opened 2018-10-16
Minor fixes

FredrikFolkesson:
* Adding newline to version command
* Adding a simple editorconfig
* Fixed lint errors in the README.md

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -8,3 +8,7 @@ root = true
 charset = utf-8
 end_of_line = lf
 insert_final_newline = true
+
+[*.go]
+indent_style = tab
+indent_size = 4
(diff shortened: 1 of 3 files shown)


.editorconfig
```diff
@@ -0,0 +1,10 @@
+# EditorConfig is awesome: http://EditorConfig.org
+
+# Top-most EditorConfig file
+root = true
+
+# Source files
+[*]
```
wennmo: Shouldn't we also specify the indentation here?

FredrikFolkesson: This is the same one as enigma.go, I would like to have 2 chars space indentation, but currently we seem to be using tabs?

PR #87, opened 2018-10-19
Design proposition for how the corectl command could be structured

FredrikFolkesson:
Currently there is not really a red thread or any structure with the different corectl commands.

This is a suggestion on how we could structure it along with some questions.

Please read and give feedback :)

### DIFF
--- a/command-design.md
+++ b/command-design.md
@@ -131,14 +131,16 @@ Usage:
 Available Commands:
   apps        Prints a list of all apps available in the current engine
   assoc       Print table associations summary
-  data        Evalutes the hypercube data of an object defined by the --object parameter. Note that only basic hypercubes like straight tables are supported
+  dimension   Shows content of an generic dimension
+  dimensions  Prints a list of all generic dimensions in the current app
   field       Shows content of a field
   fields      Print field list
   keys        Print key-only field list
-  layout      Evalutes the hypercube layout of an object defined by the --object parameter
+  measure     Shows content of an generic measure
+  measures    Prints a list of all generic measures in the current app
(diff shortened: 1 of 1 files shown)


command-design.md
```diff
@@ -0,0 +1,224 @@
...
+  layout      Evalutes the hypercube layout of an object defined by the --object parameter
+  meta        Shows metadata about the app
+  objects     Prints a list of all objects in the current app
+  properties  Prints the properties of the object identified by the --object flag
+  script      Print the reload script
+  status      Prints status info about the connection to engine and current app
+  tables      Print tables summary
+
+Flags:
+  -c, --config string   path/to/config.yml where parameters can be set instead of on the command line
```
carlioth: This section will need to specify app as well? 
```suggestion
    -a, --app string      App name including .qvf file ending. If no app is specified a session app is used instead.
```

command-design.md
```diff
@@ -0,0 +1,224 @@
...
+### Get Command
+
+```bash
+Lists one or several resources
+
+Usage:
+  corectl get [command]
+
+Available Commands:
+  apps        Prints a list of all apps available in the current engine
```
carlioth: Following the structure of kubectl I would say it makes sense to keep ```get apps```

command-design.md
FredrikFolkesson: No, not on the `get` level, since all commands there does not use the `app` parameter.

The `app` param would be one step `down` as in `kubectl get tables --app my-cool.qvf`

So the `app` param is a flag for the `get tables` command

FredrikFolkesson:
closed with #96

PR #91, opened 2018-11-02
[WIP] Support pushing of measures

FredrikFolkesson:
Work in progress PR, still needs tests and manual verification.

We need to decide if the "top level" of thing that we call "object" in `corectl` is an `genericObject` or if that is a level "above" IE if `genericObjects`, `generidMeasures` and `genericDimensions` are all are "objects"

This PR brings these changes:

Everything that before updated `genericObjects` (with the `--objects` flag) now also updates `genericMeasures` with the `--measures` flag.

`corectl objects` lists every "object" both `genericObject` and `genericMeasures` 
This is the same as before.

if you only want a specific type you can specify it with `--type=measure` or --`type=snapshot`

`corectl layout` and `corectl properties` can now also take the `--type` flag to specifify if the thing is an `genericObject` or an `genericMeasure`, if ommited a `genericObject` is assumed.

The file `objects.go` has been renamed `entities.go` if we decide that `object` in `corectl` does not mean `genericObject` but the "super type" that both `genericMeasures` and `genericObjects` are then I think it should be renamed back to `objects.go`


FredrikFolkesson:
closed with #96

PR #98, opened 2018-11-26
Start catwalk from corectl

FredrikFolkesson:
This PR makes it possible to open catwalk from corectl.

The specified app will be opened. If no app is specified the hub in catwalk will be opened instead.

I have not added any integration test since we are not producing any text output and are using an external lib for opening the browser.

The creating of the URL is tested in the unit tests already.

### DIFF
--- a/docs/corectl.md
+++ b/docs/corectl.md
@@ -20,7 +20,7 @@ corectl [flags]
 ### SEE ALSO
 
 * [corectl build](corectl_build.md)	 - Reloads and saves the app after updating connections, dimensions, measures, objects and the script
-* [corectl catwalk](corectl_catwalk.md)	 - Opens the specified app in Catwalk
+* [corectl catwalk](corectl_catwalk.md)	 - Opens the specified app in catwalk
 * [corectl eval](corectl_eval.md)	 - Evaluates a list of measures and dimensions
 * [corectl get](corectl_get.md)	 - Lists one or several resources
 * [corectl reload](corectl_reload.md)	 - Reloads the app.
--- a/docs/corectl.md
+++ b/docs/corectl.md
@@ -20,6 +20,7 @@ corectl [flags]
 ### SEE ALSO
 
 * [corectl build](corectl_build.md)	 - Reloads and saves the app after updating connections, dimensions, measures, objects and the script
+* [corectl catwalk](corectl_catwalk.md)	 - Opens the specified app in Catwalk
 * [corectl eval](corectl_eval.md)	 - Evaluates a list of measures and dimensions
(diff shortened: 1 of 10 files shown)


axelssonHakan:
Is the `ttl` needed?

main.go
```diff
@@ -85,6 +86,24 @@ var (
 		},
 	}
 
+	catwalkCmd = &cobra.Command{
+		Use:   "catwalk",
+		Short: "Opens the specified app in Catwalk",
+		Long: `Opens the app in Catwalk. Example: corectl catwalk --app my-app.qvf
```
wennmo: There is an `Example` section in a cobra command. Think we should start using it at least for new commands.

main.go
```diff
@@ -85,6 +86,24 @@ var (
 		},
 	}
 
+	catwalkCmd = &cobra.Command{
+		Use:   "catwalk",
+		Short: "Opens the specified app in Catwalk",
```
wennmo: Think we always refer to `catwalk` with lowercase

main.go
```diff
@@ -85,6 +86,24 @@ var (
...
+			
+`,
+		PersistentPreRun: func(ccmd *cobra.Command, args []string) {
+			corectlCommand.PersistentPreRun(corectlCommand, args)
+			viper.BindPFlag("engine", ccmd.PersistentFlags().Lookup("engine"))
+			viper.BindPFlag("ttl", ccmd.PersistentFlags().Lookup("ttl"))
+			viper.BindPFlag("app", ccmd.PersistentFlags().Lookup("app"))
+		},
+		Run: func(ccmd *cobra.Command, args []string) {
+			catwalkURL := "https://catwalk.core.qlik.com/?engine_url=" + internal.TidyUpEngineURL(viper.GetString("engine")) + "/ttl/" + viper.GetString("ttl") + "/apps/" + viper.GetString("app")
```
wennmo: Should we also make it possible to run against a local version of catwalk?

docs/corectl.md
```diff
@@ -20,6 +20,7 @@ corectl [flags]
 ### SEE ALSO
 
 * [corectl build](corectl_build.md)	 - Reloads and saves the app after updating connections, dimensions, measures, objects and the script
+* [corectl catwalk](corectl_catwalk.md)	 - Opens the specified app in Catwalk
```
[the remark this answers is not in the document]
FredrikFolkesson: I like catwalk better since that is the tool that is actually opened. But I guess UI could be better if you would not know what catwalk is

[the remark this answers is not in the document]
wennmo: I prefer catwalk as well. If naming it UI I would expect that it was part of this repo.

PR #107, opened 2018-12-20
Wait for onConnected event from engine and print event messages if in verbose mode

### LINKED ISSUE #90: Catch no license message on ws
When engine is configuered to run with a license service and the license service isn't working / no licenses left, engine will return on the websocket a message explaining this. 
Catch this error and display in the console. Today we will get: 
`websocket: close 1006 (abnormal closure): unexpected EOF`

FredrikFolkesson:
Wait for `onConnected` event from engine before moving on.

Print session events if in verbose mode

This closes #90

### DIFF
--- a/internal/state.go
+++ b/internal/state.go
@@ -37,14 +37,8 @@ func logConnectError(err error, engine string) {
 	os.Exit(1)
 }
 
-// PrepareEngineState makes sure that the app idenfied by the supplied parameters is created or opened or reconnected to
-// depending on the state. The TTL feature is used to keep the app session loaded to improve performance.
-func PrepareEngineState(ctx context.Context, engine string, appID string, ttl string, headers http.Header, createAppIfMissing bool) *State {
-	LogVerbose("---------- Connecting to app ----------")
-
+func connectToEngine(ctx context.Context, engine string, appID string, ttl string, headers http.Header) *enigma.Global {
 	engineURL := buildWebSocketURL(engine, ttl)
-	var doc *enigma.Doc
-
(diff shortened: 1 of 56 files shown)


internal/state.go
```diff
@@ -74,13 +76,12 @@ func PrepareEngineState(ctx context.Context, engine string, appID string, ttl st
...
-				fmt.Println(x.Topic, string(x.Content))
-			}
-		}
-	}()
+	sessionMessages := global.SessionMessageChannel()
+	err := waitForOnConnectedMessage(sessionMessages)
+	if err != nil {
+		FatalError("Failed to connect to engine with error message: ", err)
+	}
+	go printSessionMessagesIfInVerboseMode(sessionMessages)
```
gabbaxx: We are "looping" over the same channel twice. Doesn't this mean that the second loop will never receive the items that the first loop consumed?

internal/state.go
```diff
@@ -131,18 +132,48 @@ func PrepareEngineState(ctx context.Context, engine string, appID string, ttl st
...
+
 	if err != nil {
 		logConnectError(err, engine)
 	}
+	sessionMessages := global.SessionMessageChannel()
+	err = waitForOnConnectedMessage(sessionMessages)
+	if err != nil {
+		FatalError("Failed to connect to engine with error message: ", err)
+	}
+	go printSessionMessagesIfInVerboseMode(sessionMessages)
```
gabbaxx: Same as above

internal/state.go
FredrikFolkesson: True, will fix

PR #128, opened 2019-02-11
Add badges to README.md

wennmo:


sublibra:
👍  Is there one for code coverage as well?

wennmo:
@sublibra We don't measure coverage (yet) on corectl, but we can add a badge when we do 😉

PR #129, opened 2019-02-12
WIP: Notify users when a later release is available

### LINKED ISSUE #48: Corectl new version notification / update
Investigate 
- [ ] how we can notify users of a new version of the tool
- [ ] how we can simplify the update (or decide if current update is good enough)

wennmo:
This PR adds a check against github if a later version is available when running the tool. I added this check on `init`, but I guess we can also just have it on the `version` command.

The check is only run on binaries built with a version, so if you want to test it locally you need to build it with e.g. `go build -ldflags "-X main.version=0.2.0" main.go`.

This closes #48

### DIFF
--- a/cmd/utils.go
+++ b/cmd/utils.go
@@ -10,6 +10,7 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/spf13/cobra/doc"
 	"github.com/spf13/viper"
+	"github.com/tcnksm/go-latest"
 )
 
 var buildCmd = &cobra.Command{
@@ -114,6 +115,11 @@ var versionCmd = &cobra.Command{
 	Short: "Print the version of corectl",
 
 	Run: func(_ *cobra.Command, args []string) {
+
+		if version != "development build" {
+			checkLatestVersion()
+		}
+
 		fmt.Printf("corectl version: %s\n", version)
 	},
 }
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -963,6 +964,11 @@ corectl eval by "Region" // Returns the values for dimension "Region"`,
 
 func init() {
 
+	// If running a released version then check if there is a later version
+	if version != "" {
+		checkLatestVersion()
```
peol: I'm not too fond of us doing this call every execution of corectl. What happens if you are offline or on a slow network, will it block execution?

wennmo: I have added a check for if the github request failed. Seems to cause no obvious delay when being offline. Slow network however might be harder to verify.

wennmo:
I agree that we should do some additional testing before bringing this in for all commands. If we only perform the check if e.g. using `version` command then it will not be as obvious for a user that a new version is available.

main.go
wennmo: If we run it on a separate thread it might not be the first print in the log. Looks kind of weird imo, but maybe not a major issue.

wennmo:
Maybe it's better if we put the check before exiting `main`? Then it would be the last printout. 

Creating a file in the user home folder might be worth looking into, but then I guess we will need to have logic handling different OS. A bigger task anyway.

peol:
I think it's better to have the check at the start, but as Fredrik mentions not on every execution. Having a way of reminding the user if he/she hasn't been reminded before would be optimal I think. Also, what happens if the user dumps out JSON (will we just not output this msg)?

wennmo:
After some discussions we have decided on another approach. Will open a separate PR instead, and closing this one.

PR #130, opened 2019-02-14
Add support for nested JSON objects

FredrikFolkesson:
This means that the set objects and build command can take both `GenericObjectProperties` (as before) and `GenericObjectEntry` and change to the appropriate QIX function.


internal/entities.go
```diff
@@ -83,55 +84,72 @@ func setupEntity(ctx context.Context, doc *enigma.Doc, entityPath string, entity
...
 	case "dimension":
-		dimension, err := doc.GetDimension(ctx, props.Info.Id)
+		dimension, err := doc.GetDimension(ctx, entity.Info.Id)
 		if err == nil && dimension.Handle != 0 {
-			LogVerbose("Updating dimension " + props.Info.Id)
+			LogVerbose("Updating dimension " + entity.Info.Id)
 			err = dimension.SetPropertiesRaw(ctx, entityFileContents)
 			if err != nil {
-				FatalError("Failed to update dimension "+props.Info.Id, err)
+				FatalError("Failed to update dimension "+entity.Info.Id, err)
```
wennmo: Think we should be consistent with spaces when concatenating strings. Since we are updating the logging messages we might as well fix that now.

internal/entities.go
```diff
@@ -83,55 +84,72 @@ func setupEntity(ctx context.Context, doc *enigma.Doc, entityPath string, entity
...
+			objectID = entity.Info.Id
+		} else {
+			objectID = entity.Property.Info.Id
+			isGenericObjectEntry = true
+		}
+		object, err := doc.GetObject(ctx, objectID)
 		if err == nil && object.Handle != 0 {
-			LogVerbose("Updating object " + props.Info.Id)
-			err = object.SetPropertiesRaw(ctx, entityFileContents)
+			LogVerbose("Updating object " + objectID)
```
wennmo: Perhaps have a separate log entry when setting a full property tree?

internal/entities.go
```diff
@@ -83,55 +84,72 @@ func setupEntity(ctx context.Context, doc *enigma.Doc, entityPath string, entity
...
+				err = object.SetPropertiesRaw(ctx, entityFileContents)
+			}
 			if err != nil {
-				FatalError("Failed to update object "+props.Info.Id, err)
+				FatalError("Failed to update object "+objectID, err)
 			}
 		} else {
-			LogVerbose("Creating object " + props.Info.Id)
-			_, err = doc.CreateObjectRaw(ctx, entityFileContents)
+			LogVerbose("Creating object " + objectID)
```
wennmo: Separate log entry when setting a full property tree

PR #131, opened 2019-02-20
Isolate tests

axelssonHakan:


main.go
```diff
@@ -1030,6 +1046,10 @@ func init() {
 		command.PersistentFlags().Bool("json", false, "Prints the information in json format")
 	}
 
+	for _, command := range []*cobra.Command{removeCmd} {
+		command.PersistentFlags().Bool("suppress", false, "Suppress confirm")
```
wennmo: Maybe extend the description a little e.g. `Suppress all confirmation dialogues`

test/corectl_integration_test.go
```diff
@@ -108,70 +108,127 @@ func TestConnections(t *testing.T) {
...
 	}
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			cmd := exec.Command(binaryPath, tt.args...)
+			// Setup and teardown
+			teardownTest := setupTest(t, tt)
+			defer teardownTest(t, tt)
+
+			//
```
wennmo: Should there be a comment here? 😉

test/project1/corectl-alt.yml
```diff
@@ -0,0 +1,17 @@
+engine: localhost:9076
```
wennmo: Do we really need additional project files? Kind of confusing to have multiple configs in the same folder. Can we move the test case that requires this config to another project instead?

main.go
```diff
@@ -624,12 +629,27 @@ corectl eval by "Region" // Returns the values for dimension "Region"`,
...
+				app = args[0]
+			}
+
+			var confirmed bool
+
+			if viper.GetString("suppress") != "true" {
+				confirmed = askForConfirmation(fmt.Sprintf("Do you really want to delete the app: %s?", app))
+			}
+
+			if confirmed || viper.GetString("suppress") == "true" {
```
wennmo: Maybe its easier if we let `askForConfirmation` handle the `suppress` flag? That way we don't need to implement this logic in all methods that would require a confirm dialog later on.

PR #152, opened 2019-03-13
Add parameter for opening an app without data

wennmo:
This PR adds a parameter `--no-data` that can be passed to `get`,  `set` and `remove` commands. When set `corectl` will open a document without loading the data.


peol:
This is a potentially dangerous operation, some APIs will return empty lists and saving the app in this mode will remove the data blob. Is this thoroughly tested?

wennmo:
After a discussion we have decided to reiterate which methods this flag should be applicable to, and also look into the calls to `DoSave` from the current `set` methods. Since that will be a bigger change we will handle it in a separate PR. Closing this one

PR #153, opened 2019-03-13
Check for latest release when running the version command

### LINKED ISSUE #48: Corectl new version notification / update

wennmo:
This PR adds a check for latest release on github when running the `version` command. Will only be executed on binaries built with a `main.version`.  Will also generate a download string based on OS and executable path to update the tool.

```
➜  corectl git:(latest) go build -ldflags "-X main.version=0.2.0" -o corectl main.go
➜  corectl git:(latest) ./corectl version
-------------------------------------------------
Version 0.2.0 is not latest, you should upgrade to 0.4.0 
To download the latest version you can use this command: 
'curl --silent --location "https://github.com/qlik-oss/corectl/releases/download/v0.4.0/corectl-darwin-x86_64.tar.gz" | tar xz -C /tmp && mv /tmp/corectl /Users/mow/gits/qlik-oss/corectl'
-------------------------------------------------
corectl version: 0.2.0
```

This closes #48

### DIFF
--- a/cmd/utils.go
+++ b/cmd/utils.go
@@ -199,7 +199,7 @@ func checkLatestVersion() {
 		}
 
 		fmt.Println("-------------------------------------------------")
-		fmt.Printf("Version %s is not latest, you should upgrade to %s \n", version, res.Current)
+		fmt.Printf("There is a new version available! Please upgrade for the latest features and bug fixes. You are on %s, latest version is %s. \n", version, res.Current)
 		fmt.Printf("To download the latest version you can use this command: \n")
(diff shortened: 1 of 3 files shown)


peol:
`Version 0.2.0 is not latest, you should upgrade to 0.4.0 ` sounds a little bit bossy :)

Perhaps something like `There is a new version available! Please upgrade for the latest features and bug fixes. You are on 0.2.0, latest version is 0.4.0.`

PR #159, opened 2019-03-19
Add example config and document in README.md

### LINKED ISSUE #144: Create examples for corectl
A lot of users use examples as a way to get an overview of and understand a a tool. Provide examples for most common activities such as configuring connectors, get/set objects, etc.

wennmo:
This PR adds an example config and data set to an `examples` folder. Also added a separate markdown for describing the different properties that can be set in a configuration file.

This closes #144


docs/corectl_config.md
```diff
@@ -0,0 +1,105 @@
...
+
+This property sets the URL to the engine instance that you want `corectl` to connect to by default. Can be overriden with the `-e` or `--engine` flag.
+
+```yaml
+engine: localhost:9076 
+```
+
+### app
+
+This property describes the default app name that should be used for the commands. Typically used when working against a specific app on your engine instance. If you want to open a different app than configured in the file, you can also pass the `--app` or `-a` flag.
```
sublibra: app names without .qvf works as well. Not sure if this is a feature and should be mentioned.

wennmo: I think we should only document the preferred way, which is including the extension in the name imo.

PR #164, opened 2019-03-20
Generate api spec

axelssonHakan:
First "raw" `command` > `json`
https://562-146282949-gh.circle-artifacts.com/0/home/circleci/corectl/spec.json

Should be compared with https://github.com/qlik-trial/cli-spec/blob/master/docs/clischema.md

- What's missing and what's excessive
- Change naming


cmd/root.go
```diff
@@ -29,7 +29,7 @@ var rootCmd = &cobra.Command{
 
 	PersistentPreRun: func(ccmd *cobra.Command, args []string) {
 		// if help, version or generate-docs command, no prerun is needed.
-		if strings.Contains(ccmd.Use, "help") || ccmd.Use == "generate-docs" || ccmd.Use == "version" {
+		if strings.Contains(ccmd.Use, "help") || ccmd.Use == "generate-docs" || ccmd.Use == "generate-API-spec" || ccmd.Use == "version" {
```
wennmo: `generate-spec`?

spec.json
```diff
@@ -0,0 +1,690 @@
...
+          "Name": "ttl",
+          "Usage": "Engine session time to live in seconds",
+          "DefValue": "30"
+        }
+      }
+    },
+    "catwalk": {
+      "Use": "catwalk",
+      "Short": "Opens the specified app in catwalk",
+      "Long": "Opens the specified app in catwalk. If no app is specified the catwalk hub will be opened.",
```
wennmo: I think we shouldn't include `Long` (or `Short`) in the spec since that may lead to a lot of changes to the spec when only updating documentation for a command.

cmd/utils.go
```diff
@@ -138,11 +140,93 @@ var generateDocsCmd = &cobra.Command{
...
+
+	flags.VisitAll(flag)
+
+	return flagsJSON
+}
+
+var generateSpecCmd = &cobra.Command{
+	Use:    "generate-spec",
+	Short:  "Generate API spec based on cobra commands",
+	Long:   "Generate API spec docs based on cobra commands",
```
wennmo: Maybe just skip `Long`? 😉

.circleci/config.yml
```diff
@@ -28,6 +28,13 @@ jobs:
           command: |
             ./corectl generate-docs
             git diff --exit-code --name-only docs/
+      - run:
+          name: Check if changes has been done to the CLI API
+          command: |
+            ./corectl generate-spec > spec.json
+            git diff --exit-code --name-only spec.json
```
wennmo: Should we store the spec in the root folder? Maybe in `/docs`?

spec.json
```diff
@@ -0,0 +1,690 @@
...
+          "Name": "engine",
+          "Shorthand": "e",
+          "Usage": "URL to engine (default \"localhost:9076\")"
+        }
+      }
+    },
+    "eval": {
+      "Use": "eval",
+      "Short": "Evaluates a list of measures and dimensions",
+      "Long": "Evaluates a list of measures and dimensions. To evaluate a measure for a specific dimension use the \u003cmeasure\u003e by \u003cdimension\u003e notation. If dimensions are omitted then the eval will be evaluated over all dimensions.",
```
peol: This looks weird, I think we need to redesign the `eval` command so it uses flags for dimensions/measures instead of string parsing

cmd/root.go
axelssonHakan: Fixed

axelssonHakan:
@mikaeldanielsson Updated version:
https://581-146282949-gh.circle-artifacts.com/0/home/circleci/corectl/docs/spec.json

docs/spec.json
```diff
@@ -0,0 +1,383 @@
+{
+  "name": "corectl",
+  "info": {
+    "title": "Specification for corectl",
+    "description": "corectl contains various commands to interact with the Qlik Associative Engine. See respective command for more information",
+    "version": "v0.4.0",
+    "license": "MIT"
+  },
+  "clispec": "0.1.0",
+  "description": "corectl contains various commands to interact with the Qlik Associative Engine. See respective command for more information",
```
mikaeldanielsson: No description property allowed on root object

docs/spec.json
```diff
@@ -0,0 +1,383 @@
+{
+  "name": "corectl",
+  "info": {
+    "title": "Specification for corectl",
+    "description": "corectl contains various commands to interact with the Qlik Associative Engine. See respective command for more information",
+    "version": "v0.4.0",
```
mikaeldanielsson: Version should be a string according to semantic versioning which does not allow 'v' infront of the X.Y.Z

docs/spec.json
```diff
@@ -0,0 +1,383 @@
...
+  "description": "corectl contains various commands to interact with the Qlik Associative Engine. See respective command for more information",
+  "x-qlik-stability": "experimental",
+  "flags": {
+    "app": {
+      "alias": "a",
+      "description": "App name, if no app is specified a session app is used instead."
+    },
+    "bash": {
+      "description": "Bash flag used to adapt output to bash completion format",
+      "default": "false"
```
mikaeldanielsson: Should the flag have a parameter/argument of type bool with a default value of false? The spec do not currently have default on the flag itself but on the parameter/argument

docs/spec.json
axelssonHakan: fixed https://598-146282949-gh.circle-artifacts.com/0/home/circleci/corectl/docs/spec.json

docs/spec.json
axelssonHakan: added code to be able to remove it 😳
https://598-146282949-gh.circle-artifacts.com/0/home/circleci/corectl/docs/spec.json

docs/spec.json
axelssonHakan: yes - yes it should 😄golang doesn't have undefined. either true or false

docs/spec.json
```diff
@@ -0,0 +1,383 @@
...
+    },
+    "verbose": {
+      "alias": "v",
+      "description": "Logs extra information",
+      "default": "false"
+    }
+  },
+  "commands": {
+    "build": {
+      "short": "Reloads and saves the app after updating connections, dimensions, measures, objects and the script",
```
mikaeldanielsson: short -> description

PR #166, opened 2019-03-20
Bash completion

gabbaxx:
Bash (and Zsh) completion for corectl including lookup of available apps, connections, objects, dimensions and measures


README.md
```diff
@@ -73,6 +73,12 @@ Also check out the blog post about utilizing `corectl` and `catwalk` to build yo
 
 Usage documentation can be found [here](./docs/corectl.md).
 
+`corectl` provides auto completion of commands and flags for `bash` and `zsh`. To load completion in your shell add the following to your `~/.bashrc` or `~/.zshrc` file depending on shell.
+
+`. <(corectl completion bash)` or `. <(corectl completion zsh)`
```
axelssonHakan: Don't you need `source` infront of the command? or maybe the `.` is the same???

README.md
```diff
@@ -73,6 +73,12 @@ Also check out the blog post about utilizing `corectl` and `catwalk` to build yo
...
+
+Auto completion requires `jq` and `bash-completion` to be installed.
```
axelssonHakan: Link to jq?

cmd/root.go
```diff
@@ -77,54 +78,56 @@ func Execute(mainVersion string) {
...
-	}
+	rootCmd.PersistentFlags().String("ttl", "30", "Engine session time to live in seconds")
+	viper.BindPFlag("ttl", rootCmd.PersistentFlags().Lookup("ttl"))
 
-	for _, command := range []*cobra.Command{buildCmd, evalCmd, getCmd, reloadCmd, setCmd, removeCmd} {
-		//not binding to viper since binding a map does not seem to work.
-		command.PersistentFlags().StringToStringVar(&headersMap, "headers", nil, "Headers to use when connecting to qix engine")
-	}
+	//not binding to viper since binding a map does not seem to work.
+	rootCmd.PersistentFlags().StringToStringVar(&headersMap, "headers", nil, "Headers to use when connecting to qix engine")
```
axelssonHakan: Maybe add HTTP >> "HTTP headers ..."

cmd/utils.go
```diff
@@ -138,13 +129,44 @@ var generateDocsCmd = &cobra.Command{
 	},
 }
 
+// completionCmd generates auto completion commands
+var completionCmd = &cobra.Command{
+	Use:       "completion <shell>",
+	Hidden:    true,
```
axelssonHakan: Should this be hidden?

README.md
wennmo: Yep, works with both `source` and `.`

cmd/utils.go
wennmo: I guess it depends on if we want it visible in the help or not. IMO it's enough to document it in the README and keep it as hidden.

README.md
peol: Also, perhaps move this above the source commands, feels weird to list the dependencies _after_ they started installing the script

cmd/root.go
```diff
@@ -21,11 +21,12 @@ var rootCtx = context.Background()
...
 var rootCmd = &cobra.Command{
-	Hidden:            true,
-	Use:               "corectl",
-	Short:             "",
-	Long:              `Corectl contains various commands to interact with the Qlik Associative Engine. See respective command for more information`,
-	DisableAutoGenTag: true,
+	Hidden:                 true,
+	Use:                    "corectl",
+	Short:                  "",
+	Long:                   `Corectl contains various commands to interact with the Qlik Associative Engine. See respective command for more information`,
```
peol: Isn't corectl lowercase?

cmd/root.go
peol: Qlik Associative Engine

cmd/root.go
```diff
@@ -77,54 +78,56 @@ func Execute(mainVersion string) {
...
-		//not binding to viper since binding a map does not seem to work.
-		command.PersistentFlags().StringToStringVar(&headersMap, "headers", nil, "Headers to use when connecting to qix engine")
-	}
+	//not binding to viper since binding a map does not seem to work.
+	rootCmd.PersistentFlags().StringToStringVar(&headersMap, "headers", nil, "Headers to use when connecting to qix engine")
 
-	for _, command := range []*cobra.Command{buildCmd, catwalkCmd, evalCmd, getAssociationsCmd, getConnectionsCmd, getConnectionCmd, getDimensionsCmd, getDimensionCmd, getFieldsCmd, getKeysCmd, getFieldCmd, getMeasuresCmd, getMeasureCmd, getMetaCmd, getObjectsCmd, getObjectCmd, getScriptCmd, getStatusCmd, getTablesCmd, reloadCmd, removeCmd, setCmd} {
-		command.PersistentFlags().StringP("app", "a", "", "App name, if no app is specified a session app is used instead.")
-	}
+	rootCmd.PersistentFlags().StringP("app", "a", "", "App name, if no app is specified a session app is used instead.")
```
peol: App id, it's not the "name" in non-desktop mode, it's the guid of the app.

cmd/utils.go
peol: If it's a "public" API I think it should be visible

cmd/utils.go
```diff
@@ -138,13 +129,44 @@ var generateDocsCmd = &cobra.Command{
...
+	ValidArgs: []string{"zsh", "bash"},
+	Args:      cobra.MinimumNArgs(1),
+	Short:     "Generates auto completion scripts",
+	Long: `To load completion add the following to your ~/.bashrc or ~/.zshrc file
```
peol: Can we detect the shell and output the info for their current shell maybe?

cmd/utils.go
```diff
@@ -138,13 +129,44 @@ var generateDocsCmd = &cobra.Command{
...
+
+. <(corectl completion zsh)
```
peol: This requires corectl to be in PATH, right? Do we need a note on that in README as a requirement?

cmd/utils.go
```diff
@@ -138,13 +129,44 @@ var generateDocsCmd = &cobra.Command{
...
+
+or
+
+. <(corectl completion bash)
+
+If you are running on OSX you will also need to install bash-completion e.g through brew.
```
peol: Redundant information, I understand that we want this to be accessible but it's easy to miss one place if we need to update the info, perhaps link to a section in README?

cmd/utils.go
```diff
@@ -205,3 +227,151 @@ func checkLatestVersion() {
 		fmt.Println("\n-------------------------------------------------")
 	}
 }
+
+// Code for generating zsh bash completion script
+// Inspired by https://github.com/kubernetes/kubernetes/blob/master/pkg/kubectl/cmd/completion/completion.go
```
peol: I think we should have this similar to how we did in core-scaling: https://github.com/qlik-oss/core-scaling/blob/master/LICENSE

docs/corectl.md
```diff
@@ -13,9 +13,14 @@ corectl [flags]
 ### Options
```
peol: It says options here, but `corectl [flags]` in the instruction above, normalize?

wennmo:
I will merge this PR since we have other tasks pending on this. Lets handle any outstanding issues as separate tasks.

Issue #162, opened 2019-03-20
Consider aligning the connections parameter format with dimensions, measures and objects


gabbaxx:
The connections parameter differs slightly from the rest of the "list" parameters. This should ideally be aligned.

gabbaxx:
Since connections are more related to the reload script than to the entities it will stay the way it is for now.

PR #169, opened 2019-03-21
Make it possible to remove multiple connections

wennmo:
Earlier it was only possible to remove one connection at a time. To align with the other remove commands it should be possible to send in multiple IDs at once.

This is related to #162


wennmo:
Maybe we should also add aliases for these commands e.g. both `corectl remove connection ID-1` and `corectl remove connections ID-1 ID-2` triggers the same command?

PR #184, opened 2019-03-28
Completion on bash

gabbaxx:
Fixed custom completions in bash
Removed special file completions on windows (default behaviour works ok)


cmd/root.go
```diff
@@ -249,47 +241,68 @@ const bashCompletionFunc = `
...
-				COMPREPLY+=( $( compgen -W "" -- "$cur" ) )
-		fi
-  }
-
 	__corectl_get_dimensions()
 	{
 		local flags=$(__extract_flags_to_forward ${words[@]})
 		local corectl_out=$(corectl get dimensions --bash $flags 2>/dev/null)
-		__corectl_render_compreply "${corectl_out[*]}"
+		errorcode=$?
```
wennmo: Can we move this to a common function for all commands?

cmd/root.go
```diff
@@ -156,8 +145,11 @@ func init() {
 	for _, command := range []*cobra.Command{removeCmd} {
 		command.PersistentFlags().Bool("suppress", false, "Suppress all confirmation dialogues")
 	}
-
 	catwalkCmd.PersistentFlags().String("catwalk-url", "https://catwalk.core.qlik.com", "Url to an instance of catwalk, if not provided the qlik one will be used.")
+
+	if runtime.GOOS != "windows" {
```
wennmo: Maybe add a comment to why this is not applicable for windows

PR #186, opened 2019-03-29
Add possibility to open app without data

### LINKED ISSUE #163: Add support for opening app without data
There is support in `enigma-go` for opening an app without data, but it is not being used in `corectl` today. If we want to add the support we will need to investigate in what cases we want to save the app, or if we only want to save e.g. objects and not data. This to prevent any users unintentionally removing data from an app when using the no-data option.

wennmo:
This PR adds a flag `--no-data` that can be used to open an app without data. If the flag is used then only objects will be saved to the app.

When using `--no-data` or the `--ttl` flag then we do not want to attach to an existing session if the values are != default, so I have added the flags to the session id as well.

This closes #163

### DIFF
--- a/test/corectl_integration_test.go
+++ b/test/corectl_integration_test.go
@@ -20,7 +20,7 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-var update = flag.Bool("update", true, "update golden files")
+var update = flag.Bool("update", false, "update golden files")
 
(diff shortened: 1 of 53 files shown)


test/corectl_integration_test.go
```diff
@@ -20,7 +20,7 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-var update = flag.Bool("update", true, "update golden files")
+var update = flag.Bool("update", false, "update golden files")
```
axelssonHakan: Nice catch 👍

PR #203, opened 2019-04-08
Config validation

### LINKED ISSUE #198: Validation of config files
We currently do not validate property names in the config file.

A use case where this would be nice is for example `corectl.yml` file with the property `connection` instead of `connections`. The `connection` property is not used and no connections are thus created.

It might be a good idea to check for - and even prohibit - unknown properties, so that we can enforce correct configurations.

glooms:
Refactored the logic of reading config files from `cmd/root.go` into `internal/config.go`.

`ValidateConfigFile` in `internal/config.go` now reads the specified config file and uses two helper methods to do this. Incorrect properties in a config file will result in a fatal error.

Also added tests to see if validation is working correctly.
Note that all tests using config files also implicitly test config validation.

This resolves #198


internal/config.go
```diff
@@ -59,3 +59,77 @@ func ReadConnectionsFile(path string) ConnectionsConfigFile {
...
+				FatalError("Failed to read config file " + configFile)
+			}
+		}
+	}
+	QliVerbose = viper.GetBool("verbose")
+	LogTraffic = viper.GetBool("traffic")
+	if configFile != "" {
+		LogVerbose("Using config file: " + configFile)
+	} else {
+		LogVerbose("No config file")
```
axelssonHakan: Maybe add something more in the end?
- No config file found/defined???

internal/config.go
```diff
@@ -59,3 +59,77 @@ func ReadConnectionsFile(path string) ConnectionsConfigFile {
...
+
+}
+
+// validateProps reads a config file by 
+func validateProps(configPath string) {
+	source, err := ioutil.ReadFile(configPath)
+	if err != nil {
+		FatalError("Could not find config file:", configPath)
+	}
+	validProps := map[string]struct{}{ // This "set" contains the valid property names
```
peol: Drive-by review, hope you don't mind :)

Can we get this list from somewhere? I'm afraid this might get out of sync easily when we add new properties?

LGTM otherwise, good work!

peol: Also, a fun idea for another time perhaps. It's common to find recommendations in CLI for typos, e.g. `trafic` and `traffic`: `Did you mean 'traffic'`? You can easily calculate that using [Levenshtein distance](https://en.wikipedia.org/wiki/Levenshtein_distance) :)

glooms: Good point. I was thinking of solving this solution using `viper.AllSettings()`, but, this method apparently doesn't return all possible properties.

Gabriel suggested to solve this problem following the merge of his noun-verb branch where he has also collected the initialization of all flags and props to one file. From this file we can easily create the set of valid properties.

So maybe we could use this as a temporary solution and I can fix it properly following the merge of Gabriel's branch.

glooms: And yes, that would be a nice feature, but probably beyond the scope of this PR ^^
I will add it as an enhancement issue!

Issue #204, opened 2019-04-08
Suggestions when encountering typos


glooms:
Also, a fun idea for another time perhaps. It's common to find recommendations in CLI for typos, e.g. `trafic` and `traffic`: `Did you mean 'traffic'`? You can easily calculate that using [Levenshtein distance](https://en.wikipedia.org/wiki/Levenshtein_distance) :)

_Originally posted by @peol in https://github.com/qlik-oss/corectl/pull/203_

PR #205, opened 2019-04-09
Suggestion typos

glooms:
Related to the config validation we now provide some feedback in the form of suggestions for property names that are likely typos and more thorough feedback for erroneous configs in general.

Exactly what Levenshtein distance between two words we define as a typo is up for debate. I made the arbitrary choice of distance 4 as limit, using the default option in the Levenshtein library which weighs inserts and deletions as 1 and substitutions as 2.

The feedback also references the yaml's absolute path.

This close the issue #204

### DIFF
--- a/internal/config.go
+++ b/internal/config.go
@@ -8,6 +8,7 @@ import (
 	"gopkg.in/yaml.v2"
 	"github.com/spf13/viper"
 	leven "github.com/texttheater/golang-levenshtein/levenshtein"
+	"path/filepath"
 )
 
 // ConnectionConfigEntry defines the content of a connection in either the project config yml file or a connections yml file.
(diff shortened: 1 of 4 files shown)


internal/config.go
```diff
@@ -110,24 +115,65 @@ func validateProps(configPath string) {
...
+			}
 		}
 	}
-	if len(invalidProps) > 0 {
-		errorMessage := fmt.Sprintf("Found invalid config properties: %v", invalidProps)
-		FatalError(errorMessage)
+	if len(invalidProps) + len(suggestions) > 0 {
+		errorMessage := []string{}
+		errorMessage = append(errorMessage,
+			fmt.Sprintf("Corectl found invalid properties when validating the config file '%s'.", configPath))
```
wennmo: Think it should be lowercase `corectl`

internal/config.go
```diff
@@ -88,7 +93,7 @@ func ReadConfigFile(explicitConfigFile string) {
 	if configFile != "" {
 		LogVerbose("Using config file: " + configFile)
 	} else {
-		LogVerbose("No config file")
+		LogVerbose("No config file specified, using default values.")
```
axelssonHakan: 😄

internal/config.go
glooms: Done

PR #209, opened 2019-04-11
Added titles to object/dimension/measure printouts

gabbaxx:


cmd/dimension.go
```diff
@@ -63,20 +64,26 @@ var listDimensionsCmd = &cobra.Command{
 	Example: "corectl dimension ls",
 
 	Run: func(ccmd *cobra.Command, args []string) {
-		listEntities(ccmd, args, "dimension", !viper.GetBool("bash"))
+		state := internal.PrepareEngineState(rootCtx, headers, false)
+		items := internal.ListDimensions(state.Ctx, state.Doc)
+		if viper.GetBool("bash") {
```
wennmo: In my opinion we check the consistency here. For other printers I think we fetch the flags in the printer to now which format to print. Here we pass it to the printer. I am fine either way as long as we are consistent.

cmd/dimension.go
```diff
@@ -63,20 +64,26 @@ var listDimensionsCmd = &cobra.Command{
...
 	Short:   "Print the properties of the generic dimension",
 	Long:    "Print the properties of the generic dimension",
 	Example: "corectl dimension properties DIMENSION-ID",
 
 	Run: func(ccmd *cobra.Command, args []string) {
-		getEntityProperties(ccmd, args, "dimension")
+		getEntityProperties(ccmd, args, "dimension", viper.GetBool("minimum"))
 	},
-}
+}, "minimum")
```
wennmo: Should this be a string?

docs/spec.json
```diff
@@ -141,7 +141,13 @@
           "description": "Print a list of all generic dimensions in the current app"
         },
         "properties": {
-          "description": "Print the properties of the generic dimension"
+          "description": "Print the properties of the generic dimension",
+          "flags": {
+            "minimum": {
```
wennmo: Not sure if `minimum` is the correctl flag name for this, since the property tree returned is quite verbose even without the client properties for e.g. objects. Perhaps `--engine-only`?

internal/entities.go
```diff
@@ -12,6 +12,18 @@ import (
 	"gopkg.in/yaml.v2"
 )
 
+type (
+	ParsedEntityListData struct {
+		Title string `json:"title"`
+	}
+
+	NamedItemWithType struct {
+		Title string `json:"title"`
```
wennmo: Pehaps there are more properties that would be interesting in these structs. `title` is of interest but for a dimension I guess which fields that are used would also be good to output, and e.g. expression for a measure. This would require separate structs and more refactoring though.

test/corectl_integration_test.go
```diff
@@ -223,7 +223,7 @@ func TestCorectl(t *testing.T) {
 		{"project 1 - get objects", defaultConnectString1, []string{"object", "ls"}, []string{"golden", "project1-objects.golden"}, initTest{true, true}},
 		{"project 1 - get object data", defaultConnectString1, []string{"object", "data", "my-hypercube"}, []string{"golden", "project1-data.golden"}, initTest{true, true}},
 		{"project 1 - get object properties", defaultConnectString1, []string{"object", "properties", "my-hypercube"}, []string{"golden", "project1-properties.golden"}, initTest{true, true}},
-		{"project 1 - get measures 1 as json", defaultConnectString1, []string{"measure", "ls"}, []string{"golden", "project1-measures-1-json.golden"}, initTest{true, true}},
+		{"project 1 - get measures 1", defaultConnectString1, []string{"measure", "ls"}, []string{"golden", "project1-measures-1.golden"}, initTest{true, true}},
```
wennmo: We should add a test for `minimum` flag as well if that is the flag name we decide upon

Issue #206, opened 2019-04-09
Allow environment variables in config file


glooms:
Closely linked to #197.
We should probably support the use of environment variables for all properties in the config file.

Issue #197, opened 2019-04-03
How should user credentials be provided when setting up connections?


glooms:
When setting up a connection to e.g. a database it might look like:
```
mysql:
  connectionstring: ...
  username: user
  password: password
  settings: ...
```
We can apparently not include environment variables in the config file, meaning that sensitive information such as username and password for database connections have to be stated in plain text.

How should this be done?

glooms:
Update: We can include passwords as environment variables (but only passwords) and this is poorly documented, if at all.

PR #214, opened 2019-04-12
Unmarshal fix

glooms:
Changed so that config information is retrieved from viper instead of use `yaml.Unmarshal` here and there throughout the project.

Main changes are:
- `internal.ConfigDir` stores the directory of the current config file
- The config file is read once
- `internal/config.go` handles the config files

With this change, we will be able to substitute environment variables in config files and work towards resolving: #206 and #197. (There is another PR in the chamber with that enhancement.)


glooms:
Maybe `internal.ConfigDir` should be stored in the state struct?

PR #217, opened 2019-04-15
Environment variable substitution in config files.

### LINKED ISSUE #197: How should user credentials be provided when setting up connections?

### LINKED ISSUE #206: Allow environment variables in config file

glooms:
Environment variables used in config files are now substituted, including external connection configs.

This PR closes #206 and closes #197


test/corectl_integration_test.go
```diff
@@ -199,6 +202,9 @@ func TestCorectl(t *testing.T) {
 	connectToEngineWithInccorectLicenseService := "--engine=" + *engine2IP
 	connectToEngineABAC := "--engine=" + *engine3IP
 
+	//Setup env vars for project 1
+	os.Setenv("CORECTL_TEST_CONNECT", "corectl-test-connect")
```
wennmo: Shouldn't this be `corectl-test-connector`?

Issue #224, opened 2019-04-15
Config props should be overridden by flags


glooms:
The problem:

`corectl object set` 

will instead of asking for additional arguments simply take the object property from the config.
This is not very intuitive and could easily lead to mistakes when one is only trying to get use information.

Change so that if e.g. `object` is specified in the terminal, `corectl` should not read info regarding `object` from the config.

Issue #230, opened 2019-04-16
Handle when entities are not found


glooms:
Linked to issue #220 which might've been the tip of the proverbial iceberg.

There seems to be many cases where we do not check if entities exist.
In such cases, we usually just send a request to engine without handling empty responses.

We should have some feedback in these cases.

PR #231, opened 2019-04-16
Argument restriction

### LINKED ISSUE #226: Restrict number of arguments for commands
We should restrict the number of arguments that can be supplied to each command to the number of arguments it actually requires.

glooms:
We now restrict the number of arguments passed to each command using cobra's builtin support.
This also changes the default behavior of implicitly getting things from the config. For example, the case:
```bash
corectl script set
```
is no longer allowed and `corectl` will not precede using the script specified in the config.
I am not sure this completely solves issue #224 it is a partial solution to it.

However, it closes #226


cmd/connection.go
```diff
@@ -1,30 +1,24 @@
...
 	Run: func(ccmd *cobra.Command, args []string) {
 		state := internal.PrepareEngineState(rootCtx, headers, true)
-		separateConnectionsFile := ""
-		if len(args) > 0 {
-			separateConnectionsFile = args[0]
-		}
+		separateConnectionsFile := args[0]
 		if separateConnectionsFile == "" {
-			separateConnectionsFile = getPathFlagFromConfigFile("connections")
+			internal.FatalError("Error: No connection file specified.")
```
wennmo: Do we need this check? Can't see the same check for e.g. entities

cmd/dimension.go
```diff
@@ -1,27 +1,21 @@
...
-	Use:   "set <glob-pattern-path-to-dimensions-files.json>",
-	Short: "Set or update the dimensions in the current app",
-	Long:  "Set or update the dimensions in the current app",
-	Example: `corectl dimension set
-corectl dimension set ./my-dimensions-glob-path.json`,
+	Use:     "set <glob-pattern-path-to-dimensions-files.json>",
+	Args:    cobra.ExactArgs(1),
+	Short:   "Set or update the dimensions in the current app",
+	Long:    "Set or update the dimensions in the current app",
+	Example: `corectl dimension set ./my-dimensions-glob-path.json`,
```
wennmo: I  guess all these can use "" quotes now since we are no longer supplying multiple lines

test/corectl_integration_test.go
```diff
@@ -171,11 +171,18 @@ func setupTest(t *testing.T, tt test) func(t *testing.T, tt test) {
 		if tt.initTest.teardown == true {
 			t.Log("\u001b[96m *** Teardown *** \u001b[0m")
 
-			args := append(tt.connectString, []string{"app", "rm", "--suppress"}...)
+			args := append(tt.connectString, "app", "ls")
 			cmd := exec.Command(binaryPath, args...)
+			output, err := cmd.CombinedOutput()
+			apps := []map[string]interface{}{}
+			json.Unmarshal(output, &apps)
+			appName := apps[0]["name"].(string)
```
wennmo: I think we need to find a better way to solve this. Right now we delete the first app in the app list regardless the name. Could lead to an unwanted deletion of an app for anyone running the integration tests. Maybe do a diff of the app list between setup and teardown?

glooms: Been trying some different solutions.

Checking for added apps is not rigorous, since you could have a app with the same name (e.g. from previous tests) which would then cause the tear down no to work and probably tests failing that shouldn't.

The easiest approach is probably to hardcode the name of test apps so that they all use the same name and we just check for that. This could at least be a temporary solution since I removed the functionality to use the properties specified in the config implicitly.
We could at a later stage add support for using config values explicitly for commands like
`app rm` as it seems useful. We just have to figure out how such commands should be written so that the user doesn't remove entities by mistake.

Another approach is to look at the used config and extract the app name from there. It is a workaround but arguably better than hardcoding the app name for tests.

cmd/connection.go
glooms: Well, we probably should have it for entities. Without it things like:
```bash
corectl object set ""
```
can be run and return no error what so ever. In fact it prints "Saving app... Done" but it didn't do anything. (?)

This is also somewhat linked to #230 as we ask engine to look for something that can't possibly exist.

test/corectl_integration_test.go
```diff
@@ -1,5 +1,3 @@
-// +build integration
```
wennmo: This needs to added again, probably the reason for the failing cci build

test/project2/corectl.yml
```diff
@@ -1,5 +1,5 @@
 engine: localhost:9076
-app: project2.qvf
+app: corectl_test_app.qvf
```
wennmo: Looks like some trailing whitespaces in the test config files

test/corectl_integration_test.go
wennmo: I think we should go for the hardcoded test app name for now, unless you have already started implementing reading from config. Going forward we probably will need to add tests for creating e.g. apps with special characters etc and then we can look further into this issue.

cmd/connection.go
wennmo: I agree

PR #234, opened 2019-04-17
Check if app exists before rm confirmation

glooms:
This PR changes so that before attempting to delete an app it checks whether or not such an app exists.
Mainly so that the user does not have to confirm something that can't happen.


glooms:
Related to #230

PR #236, opened 2019-04-17
Handle empty prints and entities not found

### LINKED ISSUE #220: No message for object with no data
Should probably return something like: no data found.

### LINKED ISSUE #230: Handle when entities are not found

glooms:
This PR provides user-feedback where previously there were none, as in:

`corectl dimension properties invalid-dimension`

It also gives feedback in the cases where we didn't check whether or not we had data to show as in the analysis commands `keys` and `assoc`.

Lastly, in the rare case that someone sets an empty loadscript, we will state that so is the case instead of printing the empty string (which is the actual contents of the loadscript).

Closes #230
Closes #220


glooms:
Closes #230 
Closes #220

glooms:
Added test coverage for keys.

internal/field.go
```diff
@@ -83,15 +83,27 @@ func getFieldContent(ctx context.Context, doc *enigma.Doc, fieldName string, cou
...
+	if len(layout.ListObject.DataPages) == 0 {
+		errMsg := fmt.Sprintf("Error: No field by name '%s'", fieldName)
+		FatalError(errMsg)
+	}
+
+	// Get hypercube layout
 	for _, page := range layout.ListObject.DataPages {
 		for _, row := range page.Matrix {
 			for _, cell := range row {
+				text := cell.Text
```
axelssonHakan: An ide is to show both the text and the num presentation🤔

Could be god in som cases e.g
" 30" != "30"

But this only shows the top soo maybe not that useful

PR #244, opened 2019-04-23
Json responses

### LINKED ISSUE #225: JSON responses
Some functions have JSON responses. For those, the behavior of the flags `traffic` and `verbose` should either be suppressed or altered so that the response is still in JSON-format.

glooms:
This PR adds a json flag at root level. This flag will override the verbose and traffic flag since so that any non-json output is suppressed.

Some restructuring regarding where the printing and logging functions are located is also part of this PR.

This ensures that json-responses contain only json and thus closes #225

### DIFF
--- a/cmd/flags.go
+++ b/cmd/flags.go
@@ -47,7 +47,7 @@ func initGlobalFlags(globalFlags *pflag.FlagSet) {
 	globalFlags.StringP("engine", "e", "localhost:9076", "URL to the Qlik Associative Engine")
 	globalFlags.StringP("app", "a", "", "App name, if no app is specified a session app is used instead")
 	globalFlags.String("ttl", "30", "Qlik Associative Engine session time to live in seconds")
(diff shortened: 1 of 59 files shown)


cmd/flags.go
```diff
@@ -47,6 +47,7 @@ func initGlobalFlags(globalFlags *pflag.FlagSet) {
 	globalFlags.StringP("engine", "e", "localhost:9076", "URL to the Qlik Associative Engine")
 	globalFlags.StringP("app", "a", "", "App name, if no app is specified a session app is used instead")
 	globalFlags.String("ttl", "30", "Qlik Associative Engine session time to live in seconds")
+	globalFlags.Bool("json", false, "Returns output in JSON format if possible, overrides the verbose and traffic flags")
```
axelssonHakan: Technically it disables the flags, but maybe for the user it should state that verbose and traffic output is disabled

glooms: Sounds good, updated the text now.

PR #251, opened 2019-04-25
Do not modify the engine url if a full ws url is provided

wennmo:
There were earlier problems with `corectl` appending `/app/engineData/ttl/30` to the engine url provided when e.g. connecting to Qlik Sense. I have changed so that `corectl` will not modify the url if it is prefixed with `ws://` or `wss://`. If the url is e.g. `localhost:9076` then we will add `/app/ttl/30`.

Since the base websocket url is used in multiple places e.g. when fetching meta data and opening catwalk I had to do some restructuring as well.


wennmo:
This will require some more work. We still need to support `ttl` flag for full websocket urls and perhaps also parse the appID from a websocket url since we no longer support session apps.

PR #252, opened 2019-04-26
Handle websocket urls including paths and app identifiers

### LINKED ISSUE #193: Don't default to /app/engineData
If I specify `ws://localhost:8080/sense/app/ctrl00.qvf` as engine url corectl will append /app/engineData` to the url.

This makes it very hard to work with multiple different apps, since default ttl is 30s, if we always connect to the same engine session.

wennmo:
Earlier `corectl` always appended `/app/engineData/ttl/<ttl>` to an engine url which led to incorrect behaviour when e.g. connecting to Qlik Sense Enterprise. Changed so that we only modify the url if it does not contain a path. Also if no app name is provided `corectl` will check if one is present in the url.

This closes #193

### DIFF
--- a/internal/state.go
+++ b/internal/state.go
@@ -279,7 +279,9 @@ func tryParseAppFromURL(engineURL string) string {
 	re, _ := regexp.Compile("/app/([^/]+)")
 	values := re.FindStringSubmatch(u.Path)
 	if len(values) > 0 {
-		return values[1]
+		appName := values[1]
+		LogVerbose("Found app in engine url: " + appName)
+		return appName
 	}
 	return ""
 }
--- a/internal/state.go
+++ b/internal/state.go
@@ -276,7 +276,7 @@ func getSessionID(appID string) string {
 func parseAppFromURL(engineURL string) string {
 	u, _ := neturl.Parse(engineURL)
 	// Find any string in the path succeeding "/app/", and excluding anything after "/"
-	re, _ := regexp.Compile("/app/([^,/]+)")
+	re, _ := regexp.Compile("/app/([^/]+)")
 	values := re.FindStringSubmatch(u.Path)
 	if len(values) > 0 {
 		return values[1]
--- a/internal/state.go
+++ b/internal/state.go
@@ -10,6 +10,7 @@ import (
 	neturl "net/url"
 	"os"
 	"os/user"
+	"regexp"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


internal/state_test.go
```diff
@@ -7,19 +7,34 @@ import (
...
 	assert.Equal(t, buildMetadataURL("engine:1234", "appId"), "http://engine:1234/v1/apps/appId/data/metadata")
-	assert.Equal(t, buildMetadataURL("wss://engine", "appId"), "https://engine:9076/v1/apps/appId/data/metadata")
-	assert.Equal(t, buildMetadataURL("ws://engine", "appId"), "http://engine:9076/v1/apps/appId/data/metadata")
+	assert.Equal(t, buildMetadataURL("wss://engine", "appId"), "https://engine/v1/apps/appId/data/metadata")
+	assert.Equal(t, buildMetadataURL("ws://engine", "appId"), "http://engine/v1/apps/appId/data/metadata")
 	assert.Equal(t, buildMetadataURL("wss://engine:1234", "appId"), "https://engine:1234/v1/apps/appId/data/metadata")
 	assert.Equal(t, buildMetadataURL("ws://engine:1234", "appId"), "http://engine:1234/v1/apps/appId/data/metadata")
 }
 
 func TestBuildEngineUrl(t *testing.T) {
```
glooms: Looks good! Maybe add a test case without websocket prefix but with path and see what happens.

`engine/app/hello.qvf`

Should become: `ws://engine/app/hello.qvf` with the parsed app name `hello.qvf`

internal/state.go
```diff
@@ -226,23 +231,20 @@ func PrepareEngineStateWithoutApp(ctx context.Context, headers http.Header) *Sta
...
+	return "ws://" + engine
 }
 
 func buildWebSocketURL(engine string, ttl string) string {
 	engine = TidyUpEngineURL(engine)
-	return engine + "/app/engineData/ttl/" + ttl
+	u, _ := neturl.Parse(engine)
+	// Only modify the url if it does not contain a path or ends with a "/"
+	if u.Path == "" && engine[len(engine)-1:] != "/" {
+		engine = engine + "/app/engineData/ttl/" + ttl
```
peol: Use `corectl` or something instead of `engineData`?

wennmo: Agree, I will change it to `corectl` instead

PR #262, opened 2019-04-30
Parallelized engine communication done in loops

gabbaxx:


internal/modeldata.go
```diff
@@ -3,10 +3,9 @@ package internal
 import (
 	"context"
 	"fmt"
+	"github.com/qlik-oss/enigma-go"
```
wennmo: We should allign the version of go-fmt so that we don't get formatting changes for each change regarding dependencies

internal/modeldata.go
```diff
@@ -70,22 +69,33 @@ func (m *ModelMetadata) tableByName(name string) *TableModel {
 
 func createFieldModels(ctx context.Context, doc *enigma.Doc, fieldNames []string, restMetadata *RestMetadata) []*FieldModel {
 	result := make([]*FieldModel, len(fieldNames))
-	resultChannel := make(chan *enigma.FieldDescription, len(fieldNames))
-	for _, fieldName := range fieldNames {
-		fieldDescr, err := doc.GetFieldDescription(ctx, fieldName)
-		if err != nil {
-			fmt.Println("Unexpected error", err)
```
wennmo: Use `internal.FatalError` for consistency

PR #270, opened 2019-05-09
Revised testing

gabbaxx:
Split up the one big integration test into many small focused ones.
New tests are based on an abstraction of the corectl parameters and expectations on the output.
Also made sure that expected side-effects of commands can be verified


test/toolkit/test_toolkit.go
```diff
@@ -0,0 +1,293 @@
...
+		cmd := exec.Command(binaryPath, args...)
+
+		goldenName := toGoldenFileName(t.Name())
+
+		t.Log("\u001b[35m Executing command:" + strings.Join(cmd.Args, " ") + "\u001b[0m")
+		var err error
+		output, err = cmd.CombinedOutput()
+
+		actual := string(output)
+		t.Log("\u001b[35m Output:\n" + actual)
```
glooms: Coloring shoud be disabled after logging:
`t.Log("\u001b[35m Output:\n" + actual + "\u001b[0m")`

PR #278, opened 2019-05-13
Removed experimental label

gabbaxx:


wennmo:
The API is still tagged as `experimental` with the `x-qlik-stability` property. Should that be updated as well?

PR #289, opened 2019-05-22
Update of version command

### LINKED ISSUE #288: Update corectl version command
We should update the information functionality in the version command with the curl suggestion as a last resort.

glooms:
Updating the version command to add feedback about the latest releases and that they can be found on `snap` and eventually on `brew`. The code for brew is commented out as we are not yet available on brew.

The update suggestions are based on the operating system the user is using, thoughts if this is good and should we still include how to curl it?
Also, maybe we should describe how to install the tap on brew if they haven't already?

This closes #288


cmd/utils.go
```diff
@@ -64,47 +64,56 @@ func checkLatestVersion() {
...
-
+		fmt.Println("--------------------------------------------------")
+		fmt.Printf("corectl version: %s, latest version is %s\n", version, res.Current)
+		switch runtime.GOOS {
+			//	case "darwin":
+			//		fmt.Println("To udpate to the latest version using brew just run:")
+			//		fmt.Print("\n  brew upgrade qlik-corectl\n\n")
+			//		fmt.Println("If you prefer using curl you can instead run:")
+		case "linux":
+			fmt.Println("To udpate to the latest version using snap just run:")
```
sublibra: typo: udpate

cmd/utils.go
```diff
@@ -64,47 +64,56 @@ func checkLatestVersion() {
...
+			fmt.Print("\n  snap refresh qlik-corectl\n\n")
```
sublibra: why a mix of Println and Print with \n?

cmd/utils.go
```diff
@@ -64,47 +64,56 @@ func checkLatestVersion() {
...
+			fmt.Println("If you prefer using curl you can instead run:")
```
sublibra: wording: If you prefer curl, you can run:

cmd/utils.go
glooms: The linter complains about \n in Println ^^

PR #290, opened 2019-05-24
Error alignment

glooms:
All fatal errors should now have a similar look.

Added a FatalErrorf so that the fmt package doesn't have to be imported in some cases.

Catwalk now opens hub if no app is specified.
Also if the app specified isn't found an error will be raised and catwalk will not be opened.

Added some negative tests, mostly to check if the app isn't found or not specified.
Checking all negative cases will not be done in this PR but there is definitely a need for more such tests.

Added a unit test for environment substitution in config files.

Closes #290


internal/reload.go
```diff
@@ -57,12 +57,8 @@ func Reload(ctx context.Context, doc *enigma.Doc, global *enigma.Global, silent
 		}
 	}
 
-	if err != nil {
-		fmt.Println("Error when reloading app", err)
-	}
-	if !reloadSuccessful {
-		fmt.Println("DoReload was not successful!")
-		os.Exit(1)
+	if err != nil || !reloadSuccessful {
```
wennmo: Do we want to have the same error print for theses two cases?

internal/state.go
```diff
@@ -70,11 +63,14 @@ func connectToEngine(ctx context.Context, engine string, appName string, ttl str
 }
 
 //AppExists returns wether or not an app exists
-func AppExists(ctx context.Context, engine string, appName string, headers http.Header) bool {
+func AppExists(ctx context.Context, engine string, appName string, headers http.Header) error {
```
wennmo: return both error and bool

test/corectl_integration_test.go
```diff
@@ -168,7 +168,7 @@ func TestUsingJwt(t *testing.T) {
 	p4 := p.WithParams(toolkit.Params{Engine: *toolkit.EngineJwtIP, Config: "test/projects/using-jwts/corectl.yml"})
 
 	p1.ExpectOK().ExpectIncludes("Connected without app to").Run("status")
-	p2.ExpectErrorIncludes("Please check the --engine parameter or your config file.").Run("status")
+	p2.ExpectErrorIncludes("--engine", "config", "engine is running").Run("status")
```
wennmo: Has the expected outcome changed for this test case?

test/corectl_integration_test.go
```diff
@@ -182,27 +182,72 @@ func TestHelp(t *testing.T) {
...
+
 func TestEvalOnUnknownAppl(t *testing.T) {
 	p := toolkit.Params{T: t, Engine: *toolkit.EngineStdIP, App: t.Name()}
 	p.ExpectIncludes("Could not find app: App not found (1003)").Run("eval", "count(numbers)", "by", "xyz")
 }
 
 func TestEvalOnUnknownAppEngine(t *testing.T) {
 	p := toolkit.Params{T: t, Engine: "localhost:9999", App: t.Name()}
-	p.ExpectIncludes("Please check the --engine parameter or your config file.").Run("eval", "count(numbers)", "by", "xyz")
+	p.ExpectErrorIncludes("--engine", "config", "engine is running").Run("eval", "count(numbers)", "by", "xyz")
```
wennmo: Same here

test/corectl_integration_test.go
glooms: Yes, I updated the error message

test/corectl_integration_test.go
glooms: Same here too ^^

internal/state.go
glooms: on it!

internal/reload.go
glooms: Fixed

cmd/analyze.go
```diff
@@ -122,22 +122,32 @@ var catwalkCmd = withLocalFlags(&cobra.Command{
...
+			appSpecified = true
 		} else {
-			internal.FatalError("Please provide an app that should be opened in catwalk")
+			if internal.TryParseAppFromURL(viper.GetString("engine")) != "" {
+				appSpecified = true
+			}
+			catwalkURL += "?engine_url=" + internal.TidyUpEngineURL(engineURL)
+		}
+		if appSpecified {
+			if ok, err := internal.AppExists(rootCtx, engineURL, appID, headers); ok {
```
gabbaxx: Missing a "not" here?

internal/state.go
```diff
@@ -30,17 +30,10 @@ type State struct {
...
-		fmt.Println("Error details: ", err)
-	} else {
-		fmt.Println("Could not connect to engine on " + engine + ".")
-		fmt.Println("Please check the --engine parameter or your config file.")
-		fmt.Println("Error details: ", err)
-	}
-	os.Exit(1)
+	msg := fmt.Sprintf("could not connect to engine on %s\nDetails: %s\n", engine, err)
+	msg += fmt.Sprintln("Please check if the engine is running and that the url specified")
+	msg += fmt.Sprintln("using either the --engine flag or the config file is correct.")
```
gabbaxx: Last sentence is a bit tricky, at least for Monday morning

cmd/analyze.go
glooms: Fixed!

internal/state.go
glooms: This could be because:
  the engine isn't up and running                                                                                                                              
  incorrect url in --engine flag or 'engine' property (in config)

PR #294, opened 2019-05-28
Brew/snap readme update

### LINKED ISSUE #246: Publish to brew and snap
Publish `corectl 1.0.0` to homebrew and snapcraft using the analysis from #177

glooms:
Added instructions for downloading/installing using snap and brew.

Also added the brew instructions for the version command

Finally closes #246

### DIFF
--- a/cmd/utils.go
+++ b/cmd/utils.go
@@ -79,8 +79,10 @@ func checkLatestVersion() {
 		case "darwin":
 			fmt.Println("To update to the latest version using brew just run:")
 			fmt.Print("\n  brew upgrade qlik-corectl\n\n")
-			fmt.Println("If you don't already have the qlik-oss tap you have to add the tap first with:")
+			fmt.Println("If you don't already have the qlik-oss tap you have to first add the tap with:")
 			fmt.Print("\n  brew tap qlik-oss/taps\n\n")
+			fmt.Println("And after that, you have to install it with brew by running:")
+			fmt.Print("\n  brew install qlik-corectl\n\n")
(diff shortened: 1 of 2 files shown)


cmd/utils.go
```diff
@@ -66,18 +66,22 @@ func checkLatestVersion() {
...
 		switch runtime.GOOS {
-			//	case "darwin":
-			//		fmt.Println("To update to the latest version using brew just run:")
-			//		fmt.Print("\n  brew upgrade qlik-corectl\n\n")
-			//		fmt.Println("If you prefer curl, you can run:")
+		case "darwin":
+			fmt.Println("To update to the latest version using brew just run:")
+			fmt.Print("\n  brew upgrade qlik-corectl\n\n")
+			fmt.Println("If you don't already have the qlik-oss tap you have to add the tap first with:")
+			fmt.Print("\n  brew tap qlik-oss/taps\n\n")
```
wennmo: After adding the tap you will still need to run `brew install qlik-corectl`, so that should also be added here.

cmd/utils.go
```diff
@@ -66,18 +66,22 @@ func checkLatestVersion() {
...
+			fmt.Println("If you prefer curl, you can run:")
```
hrigner: ... you can run what? :)

glooms: After the switch it prints how to download with curl

cmd/utils.go
glooms: Good catch, fixed!

PR #304, opened 2019-06-03
Add flag for connecting to QSEfW with certificates

### LINKED ISSUE #299: Add Certificates support for QSEoW Engine
Currently the only way to use this with QSEoW is with JWT it seems, but given there's no session management, you burn through the 5 session proxy limit quite quickly. Given the tool and it's usage, I see more customers wanting to use certificates vs going through the proxy anyway, and then you also don't need to worry about session limit.

wennmo:
This PR adds a flag `certificates` that can be configured to a folder containing root and client certificates. This enables connecting to a QSEfW Engine and bypassing the proxy.

This closes #299

### DIFF
--- a/docs/corectl_config.md
+++ b/docs/corectl_config.md
@@ -125,7 +125,7 @@ dimensions:
 
 ### certificates
 
-If you want to connect to a Qlik Sense Enterprise using certificates, it is possible to use the `certificates` parameter. By specifying a path to the folder containing the CA and root certificates, `corectl` will use the certificates when authenticating. `corectl` only supports connecting with `PEM` certificates and expects that `client.pem`, `client_key.pem` and `root.pem` to be present in the configured folder.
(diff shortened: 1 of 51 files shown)


docs/corectl_config.md
```diff
@@ -122,6 +123,16 @@ dimensions:
   - ./dimension-*.json
 ```
 
+### certificates
+
+If you want to connect to a Qlik Sense Enterprise using certificates, it is possible to use the `certificates` parameter. By specifying a path to the folder containing the CA and root certificates, `corectl` will use the certificates when authenticating. `corectl` only supports connecting with `PEM` certificates and expects that `client.pem`, `client_key.pem` and `root.pem` to be present in the configured folder.
```
hrigner: "and expects that ... to be" - remove 'that' or "to be" -> "are" :)

PR #305, opened 2019-06-03
Add support for variables

### LINKED ISSUE #298: Add variables
Add support for variables analogous to dimensions/measures/objects

glooms:
This PR contains commands to manage variables. It seems that there are two types of variables: Variables and GenericVariables. As we want to use GenericVariables - I am assuming - we have to use the method GetVariableByName as it is the only getter that returns said type.

Tests (golden even) pass, don't seem to have the same problems as bookmarks.

Closes #298


cmd/variable.go
```diff
@@ -0,0 +1,104 @@
...
+		printer.PrintGenericEntityProperties(state, args[0], "variable", viper.GetBool("minimum"))
+	},
+}, "minimum")
+
+var getVariableLayoutCmd = &cobra.Command{
+	Use:     "layout <variable-id>",
+	Args:    cobra.ExactArgs(1),
+	Short:   "Evaluate the layout of an generic variable",
+	Long:    "Evaluate the layout of an generic variable",
+	Example: "corectl variable layout DIMENSION-ID",
```
wennmo: `VARIABLE-ID`

cmd/dimension.go
```diff
@@ -60,7 +60,7 @@ var listDimensionsCmd = &cobra.Command{
 	Run: func(ccmd *cobra.Command, args []string) {
 		state := internal.PrepareEngineState(rootCtx, headers, false)
 		items := internal.ListDimensions(state.Ctx, state.Doc)
-		printer.PrintNamedItemsList(items, viper.GetBool("bash"))
+		printer.PrintNamedItemsList(items, viper.GetBool("bash"), false)
```
wennmo: Do we only want to print the title for variables?

cmd/variable.go
glooms: Woops

cmd/dimension.go
glooms: With the bash-flag yes, since we can only retrieve variables by name. Normally the print for bash completion only prints IDs so it had to be modified to support this case.

PR #314, opened 2019-06-11
Unbuild

gabbaxx:
First iteration of the unbuild command!


cmd/unbuild.go
```diff
@@ -0,0 +1,34 @@
...
+	"github.com/qlik-oss/corectl/internal"
+	"github.com/spf13/cobra"
+)
+
+var unbuildCmd = withLocalFlags(&cobra.Command{
+	Use:     "unbuild",
+	Args:    cobra.ExactArgs(0),
+	Short:   "Split upp an existing app into separate entities",
+	Long:    "Split upp an existing app into separate entities",
+	Example: `corectl unbuild`,
```
wennmo: We should add an `x-qlik-stability` to `experimental` on this command.

cmd/unbuild.go
```diff
@@ -0,0 +1,34 @@
...
+	Annotations: map[string]string{
+		"command_category": "build",
+	},
+
+	Run: func(ccmd *cobra.Command, args []string) {
+		ctx := rootCtx
+		outdir := ccmd.Flag("dir").Value.String()
+		state := internal.PrepareEngineState(ctx, headers, false)
```
wennmo: Maybe we should always open without data?

cmd/unbuild.go
```diff
@@ -0,0 +1,34 @@
...
+
+		if outdir == DefaultUnbuild {
+			appLayout, _ := state.Doc.GetAppLayout(ctx)
+			if appLayout.Title != "" {
+				outdir = internal.BuildRootFolderFromTitle(appLayout.Title)
+			} else {
+				outdir = "unknown-app-unbuild"
```
wennmo: Not sure how common it is to have an app without a title, but maybe we should use the app id/name instead if title is missing?

cmd/unbuild.go
```diff
@@ -0,0 +1,34 @@
...
+import (
+	"github.com/qlik-oss/corectl/internal"
+	"github.com/spf13/cobra"
+)
+
+var unbuildCmd = withLocalFlags(&cobra.Command{
+	Use:     "unbuild",
+	Args:    cobra.ExactArgs(0),
+	Short:   "Split upp an existing app into separate entities",
+	Long:    "Split upp an existing app into separate entities",
```
wennmo: We should probably add more info to the `Long` description, e.g. what entities we do export.

docs/spec.json
```diff
@@ -364,6 +364,15 @@
     "tables": {
       "description": "Print tables for the data model in an app"
     },
+    "unbuild": {
+      "description": "Split upp an existing app into separate entities",
+      "flags": {
+        "dir": {
+          "description": "Path to a the folder where the unbuilt app is exported",
+          "default": "./\u003capp name\u003e-unbuild"
```
wennmo: We need to check with regards to the spec definition what should be present here

cmd/unbuild.go
```diff
@@ -0,0 +1,48 @@
...
+	Example: `corectl unbuild`,
+	Annotations: map[string]string{
+		"command_category": "build",
+		"x-qlik-stability": "experimental",
+	},
+
+	Run: func(ccmd *cobra.Command, args []string) {
+		ctx := rootCtx
+		viper.Set("no-data", "true") // Force no-data since we only use metadata
+		outdir := ccmd.Flag("dir").Value.String()
```
glooms: Admittedly very tiny, but maybe `outdir` should be `outDir`?
Otherways, really good work!

PR #322, opened 2019-06-17
Remove unmaintained dependency go-latest

wennmo:
Dependency `go-latest` has not been maintained for a while and when renovate tries to update sub-dependencies it often fails. This PR removes `go-latest` and makes use of more notable dependencies `go-github` by Google and `go-version` by Hashicorp instead.

### DIFF
--- a/cmd/utils.go
+++ b/cmd/utils.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"bufio"
+	"context"
 	"fmt"
 	"log"
 	"os"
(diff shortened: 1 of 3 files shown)


cmd/utils.go
```diff
@@ -108,6 +108,23 @@ func checkLatestVersion() {
 	}
 }
 
+func compareVersions(currentTag string, latestTag string) (bool, string) {
```
glooms: Is there a particular reason for `(bool, string)` instead of `(string, bool)`?
I've mostly seen this sort of functions called like `response, ok := someFunc()`

cmd/utils.go
```diff
@@ -108,6 +108,23 @@ func checkLatestVersion() {
...
+	currentVersion, err := ver.NewVersion(currentTag)
+	if err != nil {
+		internal.FatalErrorf("Current version is not semantically versioned: %s", currentTag)
+	}
+
+	latestVersion, err := ver.NewVersion(latestTag[1:])
```
glooms: Is this to remove `v` from `vX.X.X`?

PR #323, opened 2019-06-17
File-extension completion for commands

glooms:
This PR includes:

* Updated completion so that commands that take specific file types get suggestions that match, when possible.
* A way to get completion for PowerShell and added instructions in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ Also check out the blog post about utilizing `corectl` and `catwalk` to build yo
 
 Usage documentation can be found [here](./docs/corectl.md).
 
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -106,6 +108,19 @@ complete -o default -F __start_corectl <myalias>
 ```
 where `<myalias>` should be substituted for whatever you wish to call `corectl`.
 
+### Windows
```
wennmo: Adding bash-completion to git bash is still applicable for Windows users as well.

glooms: True!

Issue #338, opened 2019-06-24
Perhaps unwanted behavior of knownApps lookup


glooms:
Right now, our app lookup uses what was passed in as the engine url to create a map url => apps.
This means, that if you have two engine urls, say `localhost:9076` and `ws://localhost:9076` they will treated as different engines by the knownApps machinery.

I wonder if we want to do some url processing beforehand, e.g. extracting the host from the url, and using that as the key in knownApps instead of the unprocessed engine url.

PR #341, opened 2019-06-26
Rest refactorization

glooms:
The rest functionality is now in a separate package under internal.

Also made the URL parsing more robust using the parsing provided by go.


internal/urlparser.go
```diff
@@ -0,0 +1,107 @@
+package internal
+
+import (
+	"net/url"
+	"regexp"
+	"strings"
+)
+
+func ParseEngineURL(engine string) (*url.URL) {
```
wennmo: Seems like we always pass `viper.GetString("engine")` to this method, so maybe the method should not take any params and just be a getter?

cmd/app.go
```diff
@@ -59,7 +60,13 @@ var importAppCmd = &cobra.Command{
 
 	Run: func(ccmd *cobra.Command, args []string) {
 		appPath := args[0]
-		appID := internal.ImportApp(appPath, viper.GetString("engine"), headers)
+		engine := internal.ParseEngineURL(viper.GetString("engine"))
+		appID, appName, err := rest.ImportApp(appPath, engine, headers)
+		if err != nil {
+			internal.FatalError(err)
+		}
+		// TODO: Do we want to parse to map to host specifically or just the engine property?
```
wennmo: This is a good point, I think we should pass the parsed host instead. We should also make this change now (if we want it) cause it might lead to problems with existing `knownApps.yml` files e.g. not mapping IDs to app name of known hosts.

internal/rest/caller.go
```diff
@@ -0,0 +1,52 @@
...
+// an error will be returned.
+func Call(req *http.Request, result interface{}, statusCodes *map[int]bool, read parseFunction) error {
+	response, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return err
+	}
+	defer response.Body.Close()
+	ok := (*statusCodes)[response.StatusCode]
+	if !ok {
+		return fmt.Errorf("got status %d", response.StatusCode)
```
wennmo: If you encounter this error, will you know which request that the status code refers to?

internal/rest/metadata.go
```diff
@@ -1,99 +1,41 @@
...
+	result := &RestMetadata{}
+	statusCodes := &map[int]bool{
+		200: true,
 	}
-	req.Header = headers
-	response, err := http.DefaultClient.Do(req)
+	err := Call(req, result, statusCodes, json.Unmarshal)
 	if err != nil {
-		fmt.Printf("The HTTP request failed with error %s\n", err)
+		fmt.Println(url.String())
```
wennmo: Why not print this error on one line?

internal/urlparser.go
```diff
@@ -0,0 +1,107 @@
...
+		FatalError("engine URL not specified")
+	}
+	u, err := parseEngineURL(engine)
+	if err != nil {
+		FatalErrorf("could not parse engine url '%s' got error: '%s'", engine, err)
+	}
+	return u
+}
+
+// ParseEngineURL parses the engine parameter and returns an websocket URL if at all possible
```
wennmo: The tabs messes up these comments, but the content is great!

internal/urlparser_test.go
```diff
@@ -0,0 +1,139 @@
...
+import (
+	"fmt"
+	"net/url"
+	"reflect"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+// This test is more like documentation of the golang's net/url
```
wennmo: Shouldn't this test be removed when the code comments were in place?

internal/rest/caller.go
glooms: The Call function is called by e.g. ImportApp which handles the error received so what went wrong will be communicated to the user.

internal/rest/metadata.go
glooms: Good catch, the prints are superfluous!

internal/urlparser.go
glooms: I saw that too! Will fix

internal/urlparser_test.go
glooms: Yes, but these tests are similar but different. It tests the functionality of the parseEngineURL method and validates that the returned URL is correct using reflection.

cmd/app.go
glooms: Yes, we can fix it soon with #338

internal/urlparser.go
glooms: Only looks strange on github appearently, looks good in editores.

PR #362, opened 2019-07-17
Support for global contexts

### LINKED ISSUE #315: Add support for global contexts
When supporting multiple Qlik products e.g. Qlik Core, Qlik Sense Enterprise and QCS etc it would be handy to be able to create and reuse contexts without having to define them in the config file. Similar to the approach in `kubectl`

There is a WIP on this [branch](https://github.com/qlik-oss/corectl/tree/context).

Basically global contexts can be created for a deployment for flags `engine`, `certificates` and `headers`. The flags will be fetched from the global config and applied if not overridden by a local config or cli flags.

glooms:
This PR adds the ability to use global contexts.

Specifically adds the subcommand context with the commands `set, rm, ls, get, clear` to work with contexts. List of the commands:

* `set` is used to create new contexts or update existing ones.

* `rm` removes the specified context

* `ls` lists the available contexts while marking which is the currently used.

* `get` returns a more exhaustive representation of a context. Current context if nothing else is specified.

* `clear` sets the current context to an empty string.

Any command that changes the current context will inform the user by printing: `Context: <current-context>`

Furthermore, the precedence will be: cli-flags > config > context. 

Closes #315


cmd/context.go
```diff
@@ -0,0 +1,129 @@
...
+
+	Run: func(ccmd *cobra.Command, args []string) {
+		previousContext := internal.UnsetCurrentContext()
+		if previousContext != "" {
+			printer.PrintCurrentContext("")
+		}
+	},
+}
+
+var contextCmd = &cobra.Command{
```
wennmo: We should tag these commands as experimental

docs/spec.json
```diff
@@ -24,10 +24,12 @@
       "alias": "c",
       "description": "path/to/config.yml where parameters can be set instead of on the command line"
     },
+    "context": {
+      "description": "Specific context that should be used when connecting"
+    },
     "engine": {
       "alias": "e",
-      "description": "URL to the Qlik Associative Engine",
-      "default": "localhost:9076"
```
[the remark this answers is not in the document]
glooms: No, I forgot that it was automatically documented, otherwise I would argue that this is a bug and not an feature and it could therefore technically be part of a patch/minor. But, since it is part of the documentation, it would be a breaking change and we should avoid it.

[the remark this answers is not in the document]
glooms: Reverted this change.

cmd/context.go
```diff
@@ -0,0 +1,151 @@
...
+Contexts store connection information such as engine url, certificates and headers,
+similar to a config. The main difference between contexts and configs is that they
+can be used globally. Used the context subcommands to configure contexts to
+facilitate app development in environments that certificates and headers are needed.
+
+The current context is the one that is being used. You can use "context get" to
+display what is in the current context and change it by setting another context with
+"context set" or unset the current context with "context unset".
+
+Note that contexts have the lowest precedence, meaning it is the last place where
```
wennmo: Think we should skip `, meaning it is the last place where corectl looks for information.`. The sentence after describes it better imo.

glooms: Fixed!

cmd/flags.go
```diff
@@ -54,6 +54,7 @@ func initGlobalFlags(globalFlags *pflag.FlagSet) {
 	globalFlags.Bool("no-data", false, "Open app without data")
 	globalFlags.Bool("bash", false, "Bash flag used to adapt output to bash completion format")
 	globalFlags.MarkHidden("bash")
+	globalFlags.String("context", "", "Specific context that should be used when connecting")
```
gabbaxx: => "Name of context to use when connecting to the Qlik Associative Engine" or something similar?

printer/contexts.go
```diff
@@ -0,0 +1,72 @@
...
+	"sort"
+
+	"github.com/olekukonko/tablewriter"
+	"github.com/qlik-oss/corectl/internal"
+)
+
+// PrintContext prints all information in a context
+func PrintContext(name string, context *internal.Context) {
+	if context == nil {
+		fmt.Println("No current context")
```
gabbaxx: Misleading name when you specify a named context to display

cmd/context.go
```diff
@@ -0,0 +1,150 @@
...
+	Run: func(ccmd *cobra.Command, args []string) {
+		_, wasCurrent := internal.RemoveContext(args[0])
+		if wasCurrent {
+			printer.PrintCurrentContext("")
+		}
+	},
+}
+
+var getContextCmd = &cobra.Command{
+	Use:   "get",
```
gabbaxx: => get [context name]

cmd/context.go
```diff
@@ -0,0 +1,150 @@
...
+		} else {
+			name = handler.Current
+		}
+		context := handler.Get(name)
+		printer.PrintContext(name, context)
+	},
+}
+
+var updateContextCmd = withLocalFlags(&cobra.Command{
+	Use:   "update",
```
gabbaxx: update [context name]

cmd/context.go
```diff
@@ -0,0 +1,150 @@
...
+Contexts store connection information such as engine url, certificates and headers,
+similar to a config. The main difference between contexts and configs is that they
+can be used globally. Use the context subcommands to configure contexts which
+facilitate app development in environments where certificates and headers are needed.
+
+The current context is the one that is being used. You can use "context get" to
+display the contents of the current context and switch context with "context set"
+or unset the current context with "context unset".
+
+Note that contexts have the lowest precedence. This means that a e.g. an --engine flag
```
gabbaxx: Extra "a" in "This means that a e.g. an --engine flag"

printer/contexts.go
```diff
@@ -0,0 +1,72 @@
...
+		return
+	}
+	fmt.Printf("Name: %s\n", name)
```
gabbaxx: Less indentation on first line

internal/context.go
```diff
@@ -0,0 +1,274 @@
...
+	Headers      map[string]string
+	Certificates string
+	Comment      string
+}
+
+var contextFilePath = path.Join(userHomeDir(), ".corectl", "contexts.yml")
+
+func CreateContext(contextName, comment string) string {
+	if contextName == "" {
+		FatalError("\"\" is not a valid context name")
```
axelssonHakan: Maybe empty instead of ""?

internal/context.go
```diff
@@ -0,0 +1,274 @@
...
+	handler := NewContextHandler()
+	var context *Context
+	if contextName != "" {
+		handler.Exists(contextName)
+		context = handler.Get(contextName)
+	} else {
+		context = handler.GetCurrent()
+	}
+	updated := context.Update(&map[string]interface{}{
+		"engine":       viper.GetString("engine"),
```
axelssonHakan: engine == wsURL?

glooms: Yes!

printer/contexts.go
glooms: Removed all indentation.

cmd/context.go
glooms: Fixed

cmd/context.go
glooms: Removed update command

cmd/context.go
glooms: Fixed

printer/contexts.go
glooms: Fixed

internal/context.go
glooms: Changed the phrasing, should be clearer now.

internal/context.go
glooms: As engine is the name of the flag and used similarly in config it seems like the best name here as well.

cmd/flags.go
glooms: Updated to more descriptive text similar to that of engine url.

PR #372, opened 2019-07-24
Restructure logging and add quiet flag

### LINKED ISSUE #45: Investigate log verbosity and suggest a solution
Investigate how we can balance log output from different tools against cluttering the cli output.

### LINKED ISSUE #342: Quiet flag for corectl
Would be nice to have a flag `-q` that returns only a list of IDs for a command so that commands can be chained on command line, example:

```bash
> corectl connection ls -q
1234 1235 1236
> corectl connection rm $(corectl connection ls -q)
```

This feature is implemented by docker and is something I use frequently as it is very useful when working with scripts and such.

glooms:
This PR contains two things:
1. Another stab at introducing some sort of logging framework. I chose to create a new branch as the **log-verbosity** branch is 150+ commits after master.
2. A quiet flag feature.

Beginning my work on introducing a quiet flag it felt evident that we need some sort of logging framework. The result was this.

All logging handled by the `internal/log` package. It has various logging levels that follow what was suggested by @peol in #45 with the exception of quiet instead of silent and trace is omitted. Trace is for the moment unused. The information printed by the logger is prefixed with `ERROR:` for fatal and error, otherwise no prefixes are added.

This doesn't change any user-facing behavior but refactors how logging is handled internally.

The addition of the quiet flag is the only "user-facing" change in this PR.

Closes #342


glooms:
I forgot to mention that the expected behavior of the `--json` flag is a bit puzzling.
If it is set should we:

1. Ignore it where not applicable (and decide where it is applicable somehow).
2. Log according to the log level but in JSON format (currently how its implemented).
3. Keep the previous behavior --  which is similar to **1.** -- where the `--json` only applies on fatal errors and some functions.

Also, the `--json` flag overrides both the `--verbose` and `--traffic` flags, is this desired?

glooms:
One last thing, traffic could easily be logged to a file, we could simply create a `TrafficFileLogger` or something that takes a file when created and uses `fmt.Fprintln(...)` instead.

test/golden/TestReload_reload_--silent.golden
```diff
@@ -1,3 +1,3 @@
-Reload finished successfully
-Saving app... Done
-
+INFO: Reload finished successfully
+INFO: Saving app...
+INFO: App successfully saved
```
wennmo: Can we also add a test case for when passing a different log level than `INFO`?

wennmo:
@gloomns 
Regarding JSON traffic, I would atleast prefer to be able to also log it to `stdout`. I use the flag a lot when debugging error cases, and having to open a file instead feels less intuitive.

Regarding the `--json` flag, I don't mind that it has precedence over `--verbose` etc, and if I pass it I would expect only json output regardless of other settings.

cmd/completion.go
```diff
@@ -5,14 +5,15 @@ import (
...
+	"github.com/qlik-oss/corectl/internal/log"
 	"github.com/spf13/cobra"
 )
 
 // completionCmd generates auto completion commands
 var completionCmd = &cobra.Command{
 	Use:       "completion <shell>",
-	ValidArgs: []string{"zsh", "bash"},
-	Args:      cobra.MinimumNArgs(1),
+	ValidArgs: []string{"zsh", "bash", "ps"},
```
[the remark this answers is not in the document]
glooms: Sort of. `MinimumNArgs` specifies 1 or more args when in fact we want exactly one. Furthermore, the only arguments we accept are the specified args in `ValidArgs` which weren't used. Now, cobra restricts the allowed arguments to exactly one of the values in `ValidArgs`.

cmd/completion.go
```diff
@@ -39,7 +40,7 @@ Note that bash-completion is required and needs to be installed on your system.`
 		case args[0] == "ps":
 			rootCmd.GenPowerShellCompletion(os.Stdout)
 		default:
-			fmt.Printf("'%s' is not a supported shell\n", args[0])
+			log.Errorf("'%s' is not a supported shell", args[0])
```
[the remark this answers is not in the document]
glooms: Probably, but with the new changes this should never happen.

test/golden/TestReload_reload_--silent.golden
glooms: Yes!

glooms:
JSON traffic is not changed. It is still in stdout. The --json flag will now make all log messages appear in JSON format but it will respect log levels, maybe we should squelch traffic for that flag though, as previously.

cmd/root.go
```diff
@@ -53,6 +54,8 @@ var rootCmd = &cobra.Command{
 		for key, value := range headersMap {
 			headers.Set(key, value)
 		}
+		// Initiate the printers mode
+		printer.SetMode()
```
axelssonHakan: Init?

docs/spec.json
```diff
@@ -44,6 +44,11 @@
       "description": "Open app without data",
       "default": "false"
     },
+    "quiet": {
+      "alias": "q",
+      "description": "Terse logging, limited to printing IDs where possible",
```
axelssonHakan: To fancy? 🧐 Minimal instead?

internal/bookmark.go
```diff
@@ -90,13 +91,13 @@ func setBookmark(ctx context.Context, doc *enigma.Doc, bookmarkID string, raw js
 		return err
 	}
 	if bookmark.Handle != 0 {
-		LogVerbose("Updating bookmark " + bookmarkID)
+		log.Debugln("Updating bookmark " + bookmarkID)
```
axelssonHakan: I don't think this could count as debug🤓

docs/spec.json
glooms: Be it so! I shall purge that sesquipedalian word!

internal/bookmark.go
glooms: I am tempted to agree, but, the verbose level is equivalent with the new debug. This might not be ideal, since I agree that it's not a debugging message, but the other alternative would be to put it as info instead, meaning it would be default logging.

Also, to be consistent, we should then change this to Info for all similar operations e.g. objects, measures etc... which would potentially flood the terminal. (If by default enabled.)

Maybe change the log levels to: DEBUG > INFO2? > INFO instead of TRACE > DEBUG >INFO as it is now. (Trace isn't even used at this moment so it wouldn't be that strange.)
Thoughts @wennmo @gabbaxx ?

wennmo: I agree that we can skip `TRACE`, but otherwise I think we should stick to the logging conventions of Qlik. Assuming that `INFO2` was a typo? 😉

glooms:
Not sure how to handle the distinction between logging and printing. For example, the status command with verbose flag will now return "debugging" information prefixed `DEBUG` (logging) and then a line without prefix (printing) representing the status.

It looks a bit strange and I'm not sure it's ideal.

test/golden/TestReload_reload_--silent.golden
glooms: Done!

glooms:
Confirmed that transient logging is still working correctly. @gabbaxx

PR #378, opened 2019-08-12
Create corectl_config.md

alexanderpalsson:
Re-made the corectl_config.md that was accidentally deleted.


wennmo:
Also removed the step in circle ci that emptied the `docs` folder. This should not be done since not all md files are autogenerated.

PR #394, opened 2019-09-02
Add alternate states

axelssonHakan:


cmd/alternate_state.go
```diff
@@ -0,0 +1,76 @@
+package cmd
+
+import (
+	"github.com/qlik-oss/corectl/internal"
+	"github.com/qlik-oss/corectl/printer"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+)
+
+var listAlternateStateCmd = &cobra.Command{
```
wennmo: Since this returns a list of all alternate state it should probably be named `listAlternateStatesCmd`

cmd/alternate_state.go
```diff
@@ -0,0 +1,76 @@
...
+		if !viper.GetBool("no-save") {
+			internal.Save(rootCtx, state.Doc)
+		}
+	},
+}
+
+var removeAlternateStateCmd = &cobra.Command{
+	Use:     "rm <alternate-state-name>",
+	Args:    cobra.ExactArgs(1),
+	Short:   "Removes an alternate states in the current app",
```
wennmo: Since you can only remove one state this should be singular

internal/alternate_state.go
```diff
@@ -0,0 +1,36 @@
...
+	}
+}
+
+// RemoveAlternateState will remove a named alternate state in the app
+func RemoveAlternateState(ctx context.Context, doc *enigma.Doc, alternateStateName string) {
+	states := ListAlternateState(ctx, doc)
+	stateNameExists := Contains(states, alternateStateName)
+
+	if !stateNameExists {
+		FatalErrorf("no state with the name '%s' found in the app", alternateStateName)
```
wennmo: Maybe add `alternate` to this error message to make it clearer.

cmd/alternate_state.go
```diff
@@ -0,0 +1,76 @@
...
+		}
+	},
+}
+
+var alternateStateCmd = &cobra.Command{
+	Use:     "state",
+	Short:   "Explore and manage alternate states",
+	Long:    "Explore and manage alternate states",
+	Aliases: []string{"alternatestate"},
+	Annotations: map[string]string{
```
glooms: Maybe annotate with experimental?

cmd/alternate_state.go
axelssonHakan: Fixed

cmd/alternate_state.go
axelssonHakan: Fixed

internal/alternate_state.go
axelssonHakan: Fixed

cmd/alternate_state.go
axelssonHakan: Fixed

glooms:
Ironic typo there 🙈

PR #395, opened 2019-09-02
Context can now get cert from flags, with abs path

glooms:
Fixed some issues regarding absolute paths and certificates, as well as changed how the certificate flag is used. It is now used similarly to how the config flag is used.

### DIFF
--- a/internal/context.go
+++ b/internal/context.go
@@ -5,7 +5,6 @@ import (
 	"io/ioutil"
 	"os"
 	"path"
-	"path/filepath"
 	"reflect"
 	"strings"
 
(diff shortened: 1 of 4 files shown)


internal/context.go
```diff
@@ -51,6 +52,12 @@ func SetContext(contextName, comment string) string {
 		LogVerbose("Creating context: " + contextName)
 	}
 
+	// Check if certificates should be added, if so we should add them with an absolute path
+	certificates := viper.GetString("certificates")
+	if certificates != "" {
+		certificates, _ = filepath.Abs(certificates)
```
wennmo: Is this needed? Should already be an absolute path in viper.

glooms: True! Good catch. Fixed it now.

PR #416, opened 2019-09-16
Add reload limit

axelssonHakan:


test/corectl_integration_test.go
```diff
@@ -41,6 +41,8 @@ func TestReload(t *testing.T) {
 	p := toolkit.Params{T: t, Config: "test/projects/using-entities/corectl.yml", Engine: *toolkit.EngineStdIP, App: t.Name()}
 	defer p.Reset()
 	p.ExpectOK().Run("build")
+	p.ExpectIncludes("<<  3 Lines fetched").Run("build", "--limit", "3")
```
wennmo: Can we also add a test case verifying the nbr of lines fetched when not passing the `limit` flag?

axelssonHakan: Fixed

internal/reload.go
```diff
@@ -69,10 +75,29 @@ func Reload(ctx context.Context, doc *enigma.Doc, global *enigma.Global, silent
 }
 
 func logProgress(ctx context.Context, global *enigma.Global, reservedRequestID int, skipTransientLogs bool) {
+	InteractDef := enigma.InteractDef{
```
glooms: Nit: You could do `InteractDef := &enigma.InteractDef{` so that it's already a pointer. It is more consistent with the overall style of the code.

internal/reload.go
```diff
@@ -69,10 +75,29 @@ func Reload(ctx context.Context, doc *enigma.Doc, global *enigma.Global, silent
...
+		Hidden:    false,
+		Result:    0,
+		Input:     "",
+	}
+
 	progress, err := global.GetProgress(ctx, reservedRequestID)
 	if err != nil {
 		fmt.Println(err)
 	} else {
+		// If debug is set just "OK" interaction
```
glooms: Could you clarify this comment?
Maybe add a comment on to InteractDef as well.

internal/reload.go
glooms: Nit: Not sure how much of a performance sink this is but it seems like it might be an unnecessary allocation in case we are not in debug mode. Could be prevented by a rewrite so that the allocation is only done when needed, e.g. limit > 0?
(Truly nitpicking!)

axelssonHakan: Fixed

PR #417, opened 2019-09-19
Run SetProperties in parallel when updating multiple objects

### LINKED ISSUE #410: SetProperties for commands should be run in parallel
Currently we fetch all properties in parallel when doing e.g. `unbuild`, but this needs to be added for setting all properties as well to increase performance.

wennmo:
When running e.g. the `build` command creation/updating of objects were done sequentially. This PR changes so it is done in parallel instead. Since we want the output of `object ls` to be deterministic, I have also changed so the returned objects are sorted in alphabetical order on `qId`.

This closes #410

### DIFF
--- a/internal/object.go
+++ b/internal/object.go
@@ -42,7 +42,7 @@ func (o Object) validate() error {
 func ListObjects(ctx context.Context, doc *enigma.Doc) []NamedItemWithType {
 	allInfos, _ := doc.GetAllInfos(ctx)
 	unsortedResult := make(map[string]*NamedItemWithType)
-	resultInAlphabeticalOrder := []NamedItemWithType{}
+	resultInSortedOrder := []NamedItemWithType{}
 	keys := []string{}
 
 	waitChannel := make(chan *NamedItemWithType)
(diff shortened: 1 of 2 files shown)


internal/object.go
```diff
@@ -41,7 +42,8 @@ func (o Object) validate() error {
 func ListObjects(ctx context.Context, doc *enigma.Doc) []NamedItemWithType {
 	allInfos, _ := doc.GetAllInfos(ctx)
 	unsortedResult := make(map[string]*NamedItemWithType)
-	resultInOriginalOrder := []NamedItemWithType{}
+	resultInAlphabeticalOrder := []NamedItemWithType{}
```
axelssonHakan: Alphabetical 🥺
What about a numeric

internal/object.go
```diff
@@ -85,21 +88,40 @@ func SetObjects(ctx context.Context, doc *enigma.Doc, commandLineGlobPattern str
 		if err != nil {
 			FatalErrorf("could not parse file %s: %s", path, err)
 		}
+
+		// Run in parallel
+		waitChannel := make(chan error)
```
axelssonHakan: wait on what?

PR #463, opened 2019-11-27
Allow insecure connections (self signed cert)

axelssonHakan:
Rename certificates to tlsClientConfig since it will configure both:
- certificates
- InsecureSkipVerify

The tlsClientConfig will also be created by default now (possible problem?)

The path for connecting to engine is set to "/app/**engineData**/ttl/"  for allowing pass-thru for Qlik Sense Proxy. 

An `User-Agent` header `corectl/<version> (<os>)` is added to be able to track connections in ingress etc


internal/state.go
```diff
@@ -51,15 +51,12 @@ func connectToEngine(ctx context.Context, appName, ttl string, headers http.Head
 	log.Verboseln("SessionId " + headers.Get("X-Qlik-Session"))
 
 	var dialer = enigma.Dialer{}
+	dialer.TLSClientConfig = tlsClientConfig
```
glooms: Good simplification.

internal/urlparser.go
```diff
@@ -91,8 +91,8 @@ func parseEngineURL(engine string) (*url.URL, error) {
 func buildWebSocketURL(ttl string) string {
 	u := GetEngineURL()
 	// Only modify the URL path if there is no path set
-	if u.Path == "" {
-		u.Path = "/app/corectl/ttl/" + ttl
+	if u.Path == "" || u.Path == "/" {
+		u.Path = "/app/engineData/ttl/" + ttl
```
glooms: Any reason we changed the default path?

cmd/flags.go
```diff
@@ -55,6 +55,7 @@ func initGlobalFlags(globalFlags *pflag.FlagSet) {
 	globalFlags.Bool("bash", false, "Bash flag used to adapt output to bash completion format")
 	globalFlags.MarkHidden("bash")
 	globalFlags.String("context", "", "Name of the context used when connecting to Qlik Associative Engine")
+	globalFlags.BoolP("allow-insecure", "i", false, "Enabling allow-insecure will make it possible to connect using self signed certs")
```
wennmo: In my opinion we should skip the shorthand flag, since it is not recommended to ignore certificate validation (even if we are using self signed certificates. Also change `certs` -> `certificates`

docs/spec.json
```diff
@@ -9,6 +9,11 @@
   "clispec": "0.1.0",
   "x-qlik-stability": "stable",
   "flags": {
+    "allow-insecure": {
```
wennmo: Maybe also do a patch bump on the api spec version?

PR #468, opened 2019-12-10
Enable sense login

axelssonHakan:


cmd/context.go
```diff
@@ -109,6 +109,28 @@ var clearContextCmd = &cobra.Command{
 	},
 }
 
+var loginContextCmd = &cobra.Command{
```
wennmo: I guess this is very Qlik Sense Enterprise for Windows specific functionality, so only `login` might be a bit confusing.

internal/context.go
```diff
@@ -284,3 +322,90 @@ func createContextFileIfNotExist() {
...
+func getSessionCookie(tlsClientConfig *tls.Config, engineURL string, userName string, password string) string {
+	// Verify Qlik Sense URL
+	u, err := url.Parse(engineURL)
+
+	if err != nil {
+		log.Fatalln("The engineURL doesn't seem to be correct")
+	}
+
+	if u.Scheme != "https" {
+		log.Fatalln("We only support login through secure connections (HTTPS)")
```
wennmo: Think "We" should be skipped. Perhaps `Only login through secure connections (HTTPS) is supported`?

internal/context.go
```diff
@@ -284,3 +322,90 @@ func createContextFileIfNotExist() {
...
+	postResp, postErr := hc.Do(req)
+
+	if postErr != nil {
+		log.Fatalln(postErr)
+	}
+
+	setCookie := postResp.Header.Get("Set-Cookie")
+
+	if setCookie == "" {
+		log.Fatalln("We wasn't able to get the 'X-Qlik-Session' cookie, please check your password.")
```
wennmo:
```suggestion
		log.Fatalln("Not able to get the 'X-Qlik-Session' cookie, please check your password.")
```

internal/context.go
```diff
@@ -284,3 +322,90 @@ func createContextFileIfNotExist() {
...
+	}
+
+	return strings.TrimRight(strings.Fields(setCookie)[0], ";")
+}
+
+func generateXrfkey() (xrfkey string) {
```
wennmo: If we are going to use named returns it should be consistently used, which I don't think it is :)

internal/context.go
```diff
@@ -106,6 +114,36 @@ func ClearContext() string {
...
+
+	if context == nil {
+		log.Fatalf("context '%s' wasn't found.\n", contextName)
+	}
+
+	qlikSession := getSessionCookie(tlsClientConfig, context.Engine, userName, password)
+
+	if _, ok := context.Headers["cookie"]; ok {
+		// Cookie header present
+		re := regexp.MustCompile(`X-Qlik-Session=[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}`)
```
wennmo: Isn't it enough with checking if a cookie `X-Qlik-Session` is set? Think the regex adds complexity.

internal/context.go
```diff
@@ -284,3 +322,90 @@ func createContextFileIfNotExist() {
...
+
+func getSessionCookie(tlsClientConfig *tls.Config, engineURL string, userName string, password string) string {
+	// Verify Qlik Sense URL
+	u, err := url.Parse(engineURL)
+
+	if err != nil {
+		log.Fatalln("The engineURL doesn't seem to be correct")
+	}
+
+	if u.Scheme != "https" {
```
wennmo: Do we allow `wss` in the url or will that be handled earlier?

internal/context.go
```diff
@@ -284,3 +322,90 @@ func createContextFileIfNotExist() {
...
+	urlData.Set("pwd", password)
+
+	hc := http.Client{}
+	req, err := http.NewRequest("POST", loginURL.String(), strings.NewReader(urlData.Encode()))
+
+	req.PostForm = urlData
+	req.Header.Add("Content-Type", "application/x-www-form-urlencoded")
+	req.Header.Add("x-qlik-xrfkey", xrfkey)
+
+	postResp, postErr := hc.Do(req)
```
wennmo: `postErr` -> `err`

cmd/context.go
```diff
@@ -109,6 +109,28 @@ var clearContextCmd = &cobra.Command{
...
+	Use:     "login <context-name> <username> <password>",
+	Args:    cobra.RangeArgs(1, 3),
+	Short:   "Login and set cookie for the named context",
+	Long:    "Login and set cookie for the named context",
+	Example: "corectl context login",
```
wennmo: Should probably add more examples since the arguments are not mandatory

internal/context.go
axelssonHakan: Yes sir

internal/context.go
axelssonHakan: Changed

internal/context.go
axelssonHakan: Changed

internal/context.go
axelssonHakan: Not for login since that has to be done through the proxy but it´s valid for the context soo can´t be checked earlier, or?

internal/context.go
axelssonHakan: OK

internal/context.go
```diff
@@ -106,6 +114,36 @@ func ClearContext() string {
...
+		// Cookie header present
+		re := regexp.MustCompile(`X-Qlik-Session=[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}`)
+		if re.MatchString(context.Headers["cookie"]) {
+			context.Headers["cookie"] = re.ReplaceAllString(context.Headers["cookie"], qlikSession)
+		} else {
+			context.Headers["cookie"] = fmt.Sprintf("%s; %s", context.Headers["cookie"], qlikSession)
+		}
+	} else {
+		// Cookie header has to be added
+		if len(context.Headers) == 0 {
```
glooms: nit: `context.Headers == nil` is equivalent and maybe preferred.
style: prefer `map[string]string{}` where no parameters are supplied to make. (This is me nitpicking.)

internal/context.go
glooms: You can use the `internal/urlparser.go` and use the `parseEngineURL` function. Otherwise we will handle this specific url differently.
The result URL will have ws/wss as scheme which you can just set to https.

internal/context.go
```diff
@@ -284,3 +322,89 @@ func createContextFileIfNotExist() {
...
+	if err != nil {
+		log.Fatalln("The engineURL doesn't seem to be correct")
+	}
+
+	if u.Scheme != "https" {
+		log.Fatalln("Only login through secure connections (HTTPS) is supported")
+	}
+
+	// Get username
+	if len(userName) == 0 {
```
glooms: nit: equivalent with `userName == ""` (same for password).

internal/context.go
```diff
@@ -284,3 +322,89 @@ func createContextFileIfNotExist() {
...
+	if err != nil {
+		log.Fatalln(err)
+	}
+
+	// Generate xrfkey
+	xrfkey := generateXrfkey()
+
+	//
+	loginURL := resp.Request.URL
+	q, _ := url.ParseQuery(loginURL.RawQuery)
```
glooms: nit: I think you can just do a `url.Query()` here.

internal/context.go
```diff
@@ -284,3 +322,89 @@ func createContextFileIfNotExist() {
...
+	q.Add("xrfkey", xrfkey)
+	loginURL.RawQuery = q.Encode()
+
+	urlData := url.Values{}
+	urlData.Set("username", userName)
+	urlData.Set("pwd", password)
+
+	hc := http.Client{}
+	req, err := http.NewRequest("POST", loginURL.String(), strings.NewReader(urlData.Encode()))
```
glooms: Check the err, just in case.

cmd/context.go
axelssonHakan: Let us keep login

internal/context.go
axelssonHakan: Only a novice would complain 🤔🧙‍♂️

wennmo: We are all novices when it comes to regex

cmd/context.go
```diff
@@ -109,6 +109,28 @@ var clearContextCmd = &cobra.Command{
 	},
 }
 
+var loginContextCmd = &cobra.Command{
+	Use:     "login <context-name> <username> <password>",
```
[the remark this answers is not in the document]
axelssonHakan: Added that if no content name is added the current context is used

internal/context.go
axelssonHakan: fixed

internal/context.go
axelssonHakan: Gracias

internal/context.go
axelssonHakan: `Query not declared by package url`

cmd/context.go
axelssonHakan: Only contextName is an argument now. user and password is localflags

cmd/context.go
```diff
@@ -109,6 +109,27 @@ var clearContextCmd = &cobra.Command{
...
+	corectl context login context-name`,
+
+	Run: func(ccmd *cobra.Command, args []string) {
+		var contextName string
+		switch len(args) {
+		case 1:
+			contextName = args[0]
+		case 0:
+			contextName = ""
+		}
```
glooms: You don't need the switch anymore :)

cmd/flags.go
```diff
@@ -94,6 +94,8 @@ func initLocalFlags() {
 	localFlags.Bool("minimum", false, "Only print properties required by engine")
 	localFlags.String("comment", "", "Comment for the context")
 	localFlags.BoolP("quiet", "q", false, "Only print IDs. Useful for scripting")
+	localFlags.String("user", "", "Username to be used when login in to Qlik Sense Enterprise")
```
glooms:
```suggestion
	localFlags.String("user", "", "Username to be used when logging in to Qlik Sense Enterprise")
```

cmd/flags.go
```diff
@@ -94,6 +94,8 @@ func initLocalFlags() {
...
+	localFlags.String("password", "", "Password to be used when login in to Qlik Sense Enterprise (use with caution)")
```
glooms:
```suggestion
	localFlags.String("password", "", "Password to be used when logging in to Qlik Sense Enterprise (use with caution)")
```

internal/context.go
```diff
@@ -106,6 +114,45 @@ func ClearContext() string {
...
+	var context *Context
+
+	if contextName == "" {
+		context = handler.GetCurrent()
+	} else {
+		context = handler.Get(contextName)
+	}
+
+	if context == nil {
+		log.Fatalf("context '%s' wasn't found.\n", contextName)
```
glooms: This error message might be confusing in the case where there is no current context. Perhaps check different errors depending on whether there is a `contextName` or not.

internal/context.go
```diff
@@ -106,6 +114,45 @@ func ClearContext() string {
...
+	}
+
+	qlikSession := getSessionCookie(tlsClientConfig, context.Engine, userName, password)
+
+	if _, ok := context.Headers["cookie"]; ok {
+		// Cookie header present
+		re := regexp.MustCompile(`X-Qlik-Session=[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}`)
```
glooms: :notes: Let it :bee: 🎶

internal/context.go
glooms: My bad, it was a sloppy comment!
I meant this one: https://golang.org/pkg/net/url/#URL.Query
```suggestion
	q := loginURL.Query()
```

glooms:
Wondering how we could test this new functionality without getting into mocks.

cmd/context.go
```diff
@@ -109,6 +109,27 @@ var clearContextCmd = &cobra.Command{
 	},
 }
 
+var loginContextCmd = withLocalFlags(&cobra.Command{
+	Use:   "login <context-name>",
+	Args:  cobra.RangeArgs(0, 1),
+	Short: "Login and set cookie for the named context",
+	Long:  "Login and set cookie for the named context",
```
wennmo: Maybe this `long` description can be a bit longer :) Should probably include that this can be used to log in through the proxy on QSEfW.

wennmo:
@glooms Yeah this will be hard to verify in a pipeline unless connecting to a QSEfW instance. In my opinion we should not introduce that kind of external dependency.

internal/context.go
axelssonHakan: Added

cmd/context.go
axelssonHakan: Changed

internal/context.go
axelssonHakan: Great input - thanks

PR #469, opened 2019-12-12
Add Makefile

glooms:
We could include a Makefile in our repository.

This is mainly to make testing a bit easier for developers but other things could be included as well.

Since most of our testing is through our integration test there might not be much of an incentive to track coverage since I don't think we can measure it since our tests are run with `os/exec`. We have very few unit tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ lint:
 	golint -set_exit_status
 
 start-deps:
-	AcceptEUAL=yes docker-compose -f test/docker-compose.yml up -d
+	ACCEPT_EULA=yes docker-compose -f test/docker-compose.yml up -d
 
 test: start-deps
 	go test ./... -tags=integration -count=1 -race
@@ -34,7 +34,7 @@ coverage: c.out
 
 example:
 	@echo "Starting engine in docker"
-	AcceptEUAL=yes docker-compose -f examples/docker-compose.yml up -d
+	ACCEPT_EULA=yes docker-compose -f examples/docker-compose.yml up -d
 	@echo "Building corectl"
 	go build
 	./corectl build --config examples/corectl.yml
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,8 @@ SHELL := /bin/bash
 BRANCH := $(shell git rev-parse --abbrev-ref HEAD)
 COMMIT := $(shell git rev-parse HEAD | cut -c1-12)
 
+ACCEPT_EULA ?= no
+
 build:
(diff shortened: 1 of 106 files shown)


Makefile
```diff
@@ -0,0 +1,24 @@
+SHELL := /bin/bash
+
+start-deps:
+	docker-compose -f test/docker-compose.yml up -d
```
wennmo: Since these are mainly used for development purposes maybe we should add `AcceptEULA=yes docker-compose...`

Makefile
```diff
@@ -0,0 +1,24 @@
+SHELL := /bin/bash
+
```
wennmo: Add a lint target as well and reuse it in CCI config?

Makefile
```diff
@@ -0,0 +1,24 @@
...
+c.out:
+	./coverage.sh
+
+coverage: c.out
+	go tool cover -html=c.out -o coverage.html
+	rm c.out
+
+example:
+	@echo "Starting engine in docker"
+	docker-compose -f examples/docker-compose.yml up -d
```
wennmo: Same here, this will only work if you have the env `AcceptEULA` set.

Makefile
```diff
@@ -0,0 +1,42 @@
...
+	@./corectl generate-docs
+	@./corectl generate-spec
+
+lint:
+	go fmt ./...
+	@$(shell which golint || go get -u golang.org/x/lint/golint)
+	golint -set_exit_status
+
+start-deps:
+	AcceptEUAL=yes docker-compose -f test/docker-compose.yml up -d
```
wennmo: typo :)

Makefile
```diff
@@ -0,0 +1,42 @@
...
+c.out:
+	./coverage.sh
+
+coverage: c.out
+	go tool cover -html=c.out -o coverage.html
+	rm c.out
+
+example:
+	@echo "Starting engine in docker"
+	AcceptEUAL=yes docker-compose -f examples/docker-compose.yml up -d
```
wennmo: typo

Makefile
```diff
@@ -0,0 +1,42 @@
...
+	@echo "Building corectl"
+	go build
+	./corectl build --config examples/corectl.yml
+
+.PHONY: test docs
```
wennmo: Do we need these phonys?

main.go
```diff
@@ -6,7 +6,9 @@ import (
 
 // version will be set with: go build -ldflags "-X main.version=X.Y.Z"
 var version = "development build"
```
wennmo: I guess we can leave `version` empty now as well

Makefile
glooms: Yes, since we have a `test` and `docs` folder we need phony so that make runs the build target regardless of whether they exist.

PR #472, opened 2019-12-17
Get full property tree

axelssonHakan:


cmd/object.go
```diff
@@ -77,7 +77,7 @@ var getObjectPropertiesCmd = withLocalFlags(&cobra.Command{
 		state := internal.PrepareEngineState(rootCtx, headers, tlsClientConfig, false, false)
 		printer.PrintGenericEntityProperties(state, args[0], "object", viper.GetBool("minimum"))
 	},
-}, "minimum")
+}, "minimum", "full")
```
glooms: What happens if both minimum and full is set?
Right now it seems like minimum will have precedence.

docs/spec.json
```diff
@@ -406,6 +406,10 @@
         "properties": {
           "description": "Print the properties of the generic object in JSON format",
           "flags": {
+            "full": {
+              "description": "Using 'GetFullPropertyTree' to retrive properties for children as well",
```
wennmo: typo `retrive`

PR #486, opened 2020-01-22
Fix error message

### LINKED ISSUE #443: Don't always add the generic error message
If the JWT isn't signed correctly, there is a connection to the server (sense) but the error message states the opposite

Error: could not connect to engine on wss://rd-hvs.rdlund.qliktech.com/han/
Details: x509: certificate signed by unknown authority
This probably means that there is no engine running on the specified url.
Check that the engine is up and that the url specified is correct.

muiAtQlik:
Adding a proper error message for x509 error

Closes #443

### DIFF
--- a/internal/state.go
+++ b/internal/state.go
@@ -34,7 +34,8 @@ func logConnectError(err error, engine string) {
 		msg += fmt.Sprintln("This probably means that you have provided either incorrect or no authorization credentials.")
 		msg += fmt.Sprintln("Check that the headers specified are correct.")
 	} else if strings.Contains(err.Error(), "x509") {
-		msg += fmt.Sprintln("This probably means that you have certificate that are not signed properly")
+		msg += fmt.Sprintln("This probably means that you have certificate that are not signed properly.")
+		msg += fmt.Sprintln("If you have a self signed certificate then the flag '--insecure' might solve the problem.")
 	} else {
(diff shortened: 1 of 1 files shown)


internal/state.go
```diff
@@ -33,6 +33,8 @@ func logConnectError(err error, engine string) {
 	if strings.Contains(err.Error(), "401") {
 		msg += fmt.Sprintln("This probably means that you have provided either incorrect or no authorization credentials.")
 		msg += fmt.Sprintln("Check that the headers specified are correct.")
+	} else if strings.Contains(err.Error(), "x509") {
+		msg += fmt.Sprintln("This probably means that you have certificate that are not signed properly")
```
axelssonHakan: Add .
Maybe add tip that "--insecure" could solve the issue?

internal/state.go
```diff
@@ -33,6 +33,9 @@ func logConnectError(err error, engine string) {
 	if strings.Contains(err.Error(), "401") {
 		msg += fmt.Sprintln("This probably means that you have provided either incorrect or no authorization credentials.")
 		msg += fmt.Sprintln("Check that the headers specified are correct.")
+	} else if strings.Contains(err.Error(), "x509") {
+		msg += fmt.Sprintln("This probably means that you have certificate that are not signed properly.")
```
glooms:
```suggestion
		msg += fmt.Sprintln("This probably means that you have certificates that are not signed properly.")
```

PR #515, opened 2020-04-01
Merge headers from flag, config and context

glooms:
Merge headers from our three configuration sources into one map, according to precedence.

Also, all keys are converted to lower-case to avoid unexpected behavior. Note however, that this is actually not enough to keep behavior **deterministic** since multiple keys translated to the same lower-case string will collide.

However, this is only a problem if multiple keys are declared in a strange fashion in the same configuration source, e.g.:
```yaml
headers:
  Foo: bla
  foo: bla
  FOO: bla
```
which is probably a rare thing.


pkg/dynconf/dyn_settings.go
```diff
@@ -50,46 +52,63 @@ func readSettings(commandFlagSet *pflag.FlagSet, withContext bool) *DynSettings
...
 		value := getFlagValue(commandFlagSet, flag)
-		(*allParams)[flag.Name] = value
-		(*commandLineParams)[flag.Name] = value
+		commandLineParams[flag.Name] = value
 	})
 
+	// Read default values
 	commandFlagSet.VisitAll(func(flag *pflag.Flag) {
-		if (*allParams)[flag.Name] == nil {
+		if _, ok := commandLineParams[flag.Name]; !ok {
```
gabbaxx, line 98: No real point in doing this check if we modify the getter down below (see comment there). Otherwise we need to check if the flag is specified in commandLineParams OR configParams or contextParams here

pkg/dynconf/dyn_settings.go
```diff
@@ -351,5 +389,24 @@ func (ds *DynSettings) GetTLSConfigFromPath(certificatesPath string) *tls.Config
 
 // Returns true if no value has been set in either the context, config file or command line for the supplied flag name
 func (ds *DynSettings) IsUsingDefaultValue(name string) bool {
-	return ds.defaultValueIsUsed[name]
+	_, ok := ds.defaultParams[name]
```
gabbaxx, line 392: This does not work anymore since the code building the defaultParams array above only takes commandline parameters into account. Maybe modify this to check if the parameter is in configParams or contextParams or commandLineParams.

pkg/dynconf/dyn_settings.go
glooms: Fixed now!

PR #537, opened 2020-04-28
Raw command with support for http redirects

gabbaxx:


pkg/commands/standard/raw.go
```diff
@@ -0,0 +1,224 @@
...
+		for i := 0; i < len(keyParts)-1; i++ {
+			keyPart := keyParts[i]
+			if jsonNode[keyPart] == nil {
+				newNode := make(map[string]interface{})
+				jsonNode[keyPart] = newNode
+				jsonNode = newNode
+			}
+		}
+		lastKey := keyParts[len(keyParts)-1]
+		switch valueType {
```
glooms, line 133: You can't send floats now right?
Maybe have number instead of int as it's called number in JSON.
Then you have to check if it can be parsed as an int or a float.

pkg/commands/standard/raw.go
```diff
@@ -0,0 +1,224 @@
...
+
+	}
+	jsonBytes, err := json.Marshal(rootJSonNode)
+	if err != nil {
+		return nil, err
+	}
+	return ioutil.NopCloser(bytes.NewReader(jsonBytes)), nil
+}
+
+//detectFileMimeType detects the content-type of the file be checking against various formats
```
glooms, line 158: nit: Typo, by not be.

pkg/commands/standard/raw_test.go
```diff
@@ -0,0 +1,47 @@
...
+
+func TestGetJsonBodyFromFlags(t *testing.T) {
+	for i, testcase := range []struct {
+		fileName         string
+		bodyString       string
+		params           map[string]string
+		expectedBody     string
+		expectedMimeType string
+	}{
+		{"raw_test_body.json", "", nil, "{\"test\": \"yes it is\"}\n", "application/json"},
```
glooms, line 20: tip: You can use raw strings ` so you don't have to escape "

pkg/commands/standard/raw_test.go
```diff
@@ -0,0 +1,47 @@
...
+		{"raw_test_body.qvf", "", nil, "<qvf content placeholder>\n", "binary/octet-stream"},
+		{"raw_test_body-image.png", "", nil, "<skip assert>", "image/png"},
+		{"", "inline-body-string", nil, "inline-body-string", "text/plain"},
+		{"", "{\"test\": \"yes it is\"}", nil, "{\"test\": \"yes it is\"}", "application/json"},
+		{"", "", map[string]string{"key": "value"}, "{\"key\":\"value\"}", "application/json"},
+		{"", "", map[string]string{"key:bool": "true"}, "{\"key\":true}", "application/json"},
+		{"", "", map[string]string{"nested.key": "value"}, "{\"nested\":{\"key\":\"value\"}}", "application/json"},
```
glooms, line 27: Since we'll be parsing user input here, maybe we want some negative test cases as well?
`key:bla` `nested.key.`

glooms, line 27: Do we handle the `nested.key:type` case?

pkg/rest/rest_caller.go
```diff
@@ -141,6 +141,87 @@ func (c *RestCaller) CallReq(req *http.Request, result interface{}) error {
...
+		req, err := http.NewRequest("GET", redirectUrl.String(), nil)
+		if err != nil {
+			return nil, err
+		}
+		return c.CallRaw(req)
+	}
+	return res, nil
+}
+
+//CallStreaming makes a rest call, follows redirects if present and streams the output to the supplied output
```
glooms, line 173: Nice solution!

pkg/rest/rest_caller.go
```diff
@@ -141,6 +141,87 @@ func (c *RestCaller) CallReq(req *http.Request, result interface{}) error {
...
+	}
+
+	if isJsonResponse(res) {
+		//We have got a json response
+		data, err := ioutil.ReadAll(res.Body)
+		if err != nil {
+			fmt.Println(output, err)
+			return err
+		}
+		if quiet { //Only print IDs
```
glooms, line 211: Can't we get quiet from the `c.PrintMode()` instead?

pkg/commands/standard/raw_test.go
gabbaxx: I know but  didn't get that to work with whitespacing or something so I gave up on that...

pkg/rest/rest_caller.go
gabbaxx: I think that will break the chaining support

glooms:
I think we can build handling of output-formats and such in another PR.

pkg/commands/standard/raw.go
```diff
@@ -0,0 +1,241 @@
...
+	return ioutil.NopCloser(bytes.NewReader(jsonBytes)), nil
+}
+
+//detectFileMimeType detects the content-type of the file by checking against various formats
+func detectFileMimeType(filePath string) string {
+	isJsonFile := strings.HasSuffix(strings.ToLower(filePath), ".json")
+	if isJsonFile {
+		return applicationJson
+	}
+	isQvfFile := strings.HasSuffix(strings.ToLower(filePath), ".qvf")
```
wennmo, line 181: We should probably also consider adding `.qvw` and `qvd` here.

pkg/commands/standard/raw.go
```diff
@@ -0,0 +1,241 @@
...
+const applicationJson = "application/json"
+const binaryOctetStream = "binary/octet-stream"
+const applicationOctetStream = "application/octet-stream"
+const textPlain = "text/plain"
+
+func CreateRawCommand() *cobra.Command {
+	command := &cobra.Command{
+		Use:     "raw <get/put/patch/post/delete> v1/url",
+		Example: "corectl raw get v1/items --query name=ImportantApp",
+		Short:   "Send Http API Request to Qlik Sense Cloud",
```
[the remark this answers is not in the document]
glooms: As it is built now it will only work against `<host>/api/v1` endpoints I think.
Although I think the correct term is Qlik Sense Cloud **editions**

pkg/commands/standard/raw.go
glooms: We can't upload qvd's I think, since the datafiles API isn't public, and use of it is discouraged, but qvw definitely.

pkg/rest/rest_caller.go
```diff
@@ -141,6 +141,87 @@ func (c *RestCaller) CallReq(req *http.Request, result interface{}) error {
...
+			fmt.Println(output, errorWithBody.Error())
+		}
+		return err
+	}
+
+	if isJsonResponse(res) {
+		//We have got a json response
+		data, err := ioutil.ReadAll(res.Body)
+		if err != nil {
+			fmt.Println(output, err)
```
glooms, line 208:
```suggestion
			fmt.Fprintln(output, err)
```

gabbaxx: Nice catch! Fixed it in a few other places as well!

pkg/commands/standard/raw.go
gabbaxx: Added  qvf, qvw, qvd and qvx. Note this is only a fast route so if we miss a type it will still work but take slightly more memory and time to determine the mime type

pkg/commands/standard/raw.go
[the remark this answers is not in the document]
gabbaxx: Changed to Qlik Sense Cloud editions

PR #543, opened 2020-05-05
fix: parse title from qMetaDef

### LINKED ISSUE #534: No sheets titles in app object ls functions
Function `app object ls` returns sheets without title.
`App object properties` for a specific sheet returns json with sheet title.

wennmo:
`Title` was never populated when listing objects. This PR changes so we populate `Title` from `qMetaDef/title`.

This closes #534.

### DIFF
--- a/internal/entities.go
+++ b/internal/entities.go
@@ -33,9 +33,12 @@ type (
 	// PropsWithTitle struct
 	PropsWithTitle struct {
 		*enigma.GenericObjectProperties
-		QMetaDef struct {
-			Title string `json:"title"`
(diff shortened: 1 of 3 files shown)


internal/entities.go
```diff
@@ -33,7 +33,9 @@ type (
 	// PropsWithTitle struct
 	PropsWithTitle struct {
 		*enigma.GenericObjectProperties
-		Title string `json:"title"`
+		QMetaDef struct {
```
glooms, line 36: Could this be its own struct instead of an anonymous one?

PR #553, opened 2020-05-26
New completion functions

gabbaxx:
Added completion functions for all engine commands and the app flag

### DIFF
--- a/pkg/boot/flags.go
+++ b/pkg/boot/flags.go
@@ -5,6 +5,7 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 	"runtime"
+	"strings"
 )
 
 func InjectGlobalFlags(command *cobra.Command, hideEngineSpecificFlags bool) {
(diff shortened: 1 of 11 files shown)


pkg/boot/flags.go
```diff
@@ -54,8 +55,49 @@ func InjectAppWebSocketFlags(command *cobra.Command, hideEngineSpecificAppComple
...
+	} else {
+		ctx, global, _ := comm.OpenGlobalSocket()
+		docList, err := global.GetDocList(ctx)
+
+		if err != nil {
+			return []string{}, cobra.ShellCompDirectiveError
+		}
+		completions := make([]string, 0)
+		for _, x := range docList {
+			if strings.HasPrefix(x.DocName, toComplete) {
```
glooms, line 95: You can filter the results based on what to complete but the completion also does this, so it's not necessary, but it doesn't hurt either.

pkg/boot/flags.go
```diff
@@ -54,8 +55,49 @@ func InjectAppWebSocketFlags(command *cobra.Command, hideEngineSpecificAppComple
...
+				completions = append(completions, x.DocName)
+			} else if strings.HasPrefix(x.DocId, toComplete) {
+				completions = append(completions, x.DocId)
+			}
+		}
+		return completions, 0
```
glooms, line 101: Should the `ShellCompDirective` be `cobra.ShellCompDirectiveNoFileComp` here as well?

pkg/commands/engine/analyze.go
```diff
@@ -117,10 +118,11 @@ func CreateGetKeysCommand() *cobra.Command {
 
 func CreateEvalCommand() *cobra.Command {
 	return &cobra.Command{
-		Use:   "eval <measure 1> [<measure 2...>] by <dimension 1> [<dimension 2...]",
-		Args:  cobra.MinimumNArgs(1),
-		Short: "Evaluate a list of measures and dimensions",
-		Long:  `Evaluate a list of measures and dimensions. To evaluate a measure for a specific dimension use the <measure> by <dimension> notation. If dimensions are omitted then the eval will be evaluated over all dimensions.`,
+		Use:               "eval <measure 1> [<measure 2...>] by <dimension 1> [<dimension 2...]",
+		Args:              cobra.MinimumNArgs(1),
+		ValidArgsFunction: listFieldsForCompletion,
```
glooms, line 123: I don't think `eval` should have this completion.
If `eval` should have completion, it has to be tailored to this command which is quite different from others.

gabbaxx: Yep lets keep that one for later

PR #556, opened 2020-06-01
Add raw comment to User-Agent

axelssonHakan:
"Hi-jacking" dynSetting to be able to append a comment to the User-Agent used to differentiate raw command


test/corectl_integration_test.go
```diff
@@ -595,3 +595,10 @@ func TestCertificatesPathNegative(t *testing.T) {
 		p.ExpectErrorIncludes("could not load client certificate").Run("context", "create", "cert-test")
 	}
 }
+
+func TestRaw(t *testing.T) {
+	p := toolkit.Params{T: t, Server: *toolkit.EngineStdIP}
+	defer p.Reset()
+
+	p.ExpectIncludes(") raw").Run("raw", "get", "v1/apps/privileges", "-v")
```
axelssonHakan, line 603: Do we want to implement regexp for better check?
`> User-Agent: corectl/dev (darwin) raw`

glooms: I think it's good enough.

PR #566, opened 2020-06-10
Seting Referer header

axelssonHakan:


pkg/boot/common_settings.go
```diff
@@ -63,6 +64,9 @@ func (c *CommonSettings) Headers() http.Header {
 	if agent := headers.Get("User-Agent"); agent == "" {
 		headers.Set("User-Agent", "corectl")
 	}
+
+	headers.Set("Referer", c.GetContextName())
```
glooms, line 68: As stated on the morning meeting, probably better to use the `--server` property instead.

axelssonHakan: It sure is - thanks

PR #575, opened 2020-07-30
Added publish and unpublish commands for generic objects

goldbergjeffrey:


internal/object.go
```diff
@@ -165,3 +172,51 @@ func setObject(ctx context.Context, doc *enigma.Doc, info *enigma.NxInfo, props
...
+
+func Publish(ctx context.Context, doc *enigma.Doc, objectID string) error {
+	object, err := doc.GetObject(ctx, objectID)
+	if err != nil {
+		return err
+	}
+	if object.Handle != 0 {
+		var layout Layout
+		var raw json.RawMessage
+		raw, err = object.GetLayoutRaw(ctx)
```
glooms, line 184: nit: You could write
```
raw, err := object.GetLayoutRaw(ctx)
```
That way you don't have to declare the `raw` variable on the line above.

internal/object.go
```diff
@@ -165,3 +172,51 @@ func setObject(ctx context.Context, doc *enigma.Doc, info *enigma.NxInfo, props
...
+		err = json.Unmarshal(raw, &layout)
+		if layout.Meta.Published {
+			log.Infoln(objectID + " is published.")
```
glooms, line 187: Maybe this could be clarified with `is already published.`?

internal/object.go
```diff
@@ -165,3 +172,51 @@ func setObject(ctx context.Context, doc *enigma.Doc, info *enigma.NxInfo, props
...
+			return fmt.Errorf("Cannot publish a published object: " + objectID)
```
glooms, line 188: style: Error messages in go should begin with lowercase letters.

pkg/commands/engine/object.go
```diff
@@ -111,6 +111,40 @@ The JSON objects can be in either the GenericObjectProperties format or the Gene
...
 
+	var publishObjectCmd = &cobra.Command{
+		Use:     "publish <object-id>",
+		Args:    cobra.ExactArgs(1),
+		Short:   "Publish a generic object, like to make a sheet public",
+		Long:    "Publish a generic object, like to make a sheet public",
+		Example: "corectl object publish OBJECT-ID -a APP-ID",
+
+		Run: func(ccmd *cobra.Command, args []string) {
+			commandLineObjects := args[0]
```
glooms, line 122: Suggestion: `objectID` might be a more precise name for this variable.

pkg/commands/engine/object.go
```diff
@@ -111,6 +111,40 @@ The JSON objects can be in either the GenericObjectProperties format or the Gene
...
+			if commandLineObjects == "" {
+				log.Fatalln("no objects specified")
+			}
+			ctx, _, doc, _ := boot.NewCommunicator(ccmd).OpenAppSocket(true)
+			internal.Publish(ctx, doc, args[0])
```
glooms, line 127: Use the variable defined instead of `args[0]` here?

internal/object.go
```diff
@@ -165,3 +172,51 @@ func setObject(ctx context.Context, doc *enigma.Doc, info *enigma.NxInfo, props
...
+			if err != nil {
+				return fmt.Errorf("Unable to publish %s with %s: %s", "object", objectID, err)
+			}
+		}
+	}
+	return nil
+}
+
+func UnPublish(ctx context.Context, doc *enigma.Doc, objectID string) error {
+	object, err := doc.GetObject(ctx, objectID)
```
glooms, line 201: Same comments as for Publish basically. :)

pkg/commands/engine/object.go
```diff
@@ -111,6 +111,40 @@ The JSON objects can be in either the GenericObjectProperties format or the Gene
...
+
+	var unPublishObjectCmd = &cobra.Command{
+		Use:     "unpublish <object-id>",
+		Args:    cobra.ExactArgs(1),
+		Short:   "Unpublish a generic object, like to make a sheet private",
+		Long:    "Unpublish a generic object, like to make a sheet private",
+		Example: "corectl object unpublish OBJECT-ID -a APP-ID",
+
+		Run: func(ccmd *cobra.Command, args []string) {
+			commandLineObjects := args[0]
```
glooms, line 139: Same comments as for the publish command.

pkg/commands/engine/object.go
```diff
@@ -111,6 +111,40 @@ The JSON objects can be in either the GenericObjectProperties format or the Gene
 		},
 	}
 
+	var publishObjectCmd = &cobra.Command{
+		Use:     "publish <object-id>",
+		Args:    cobra.ExactArgs(1),
+		Short:   "Publish a generic object, like to make a sheet public",
+		Long:    "Publish a generic object, like to make a sheet public",
+		Example: "corectl object publish OBJECT-ID -a APP-ID",
+
```
glooms, line 120: You could mark the introduced commands as experimental if you want by adding the following:
```go
   Annotations: map[string]string{
      "x-qlik-stability": "experimental",
   }
```

internal/object.go
goldbergjeffrey: interesting...I'm just used to declaring everything. BUf you are saying you don't have to do that because Go can autoassign a type to a variable?

internal/object.go
goldbergjeffrey: 👍

internal/object.go
goldbergjeffrey: okey dokey.

pkg/commands/engine/object.go
goldbergjeffrey: I don't follow, you mean in the example? or instead of args?

pkg/commands/engine/object.go
goldbergjeffrey: does the variable need to be an array because it's args or are you saying just a string?

pkg/commands/engine/object.go
goldbergjeffrey: where does that get added?

goldbergjeffrey: oh nevermind, I get it

internal/object.go
glooms: Yes, it can implicitly assign types to variables.

glooms: But to be clear, there's nothing wrong with the way you wrote it. I'm only saying it can be written more tersely.

pkg/commands/engine/object.go
glooms: Instead of `commandLineObjects` you could call the variable `objectID` :)

pkg/commands/engine/object.go
glooms: No you already have a variable that is assigned as `commandLineObjects := args[0]`, so you can use that instead of `args[0]` since they are equivalent. And the variable is a string as `args` is a `[]string`.

pkg/commands/engine/object.go
goldbergjeffrey: oh got it. I'm just lazy programmer dude. Copy/Paste press any key.

pkg/commands/engine/object.go
goldbergjeffrey: circle ci doesn't like the Annotations on the commands.

goldbergjeffrey:
@sublibra I'd like to merge this. Do you approve?

sublibra:
Yes. As long as the team is happy I approve any PR :-)

